PR #7, opened 2016-05-09
pls dont touch

hmoco:


addon/models/relationship.js
```diff
@@ -0,0 +1,7 @@
+import Model from 'ember-data/model';
+
+export default Model.extend({
+    target_type: DS.attr('string'),
```
abought: In general, try to follow JS conventions for names where possible (`targetType` vs `target_type`). Eg, the application adapter should take care of converting names from camelcase to snakecase and back, in cases where the API response does not match ember data conventions.

I wish I could say I was being pedantic, but Ember has a *lot* of magical assumptions based on how fields are named.

abought:
PR description would be helpful.

addon/models/relationship.js
samchrisinger: In particular, the way the ApplicationSerializer transforms these fields will probably break this (underscored -> camelized)

addon/serializers/application.js
```diff
@@ -32,6 +32,9 @@ export default DS.JSONAPISerializer.extend({
             return Ember.String.camelize(key);
         }
     },
+    keyForRelationship(key) {
```
samchrisinger: This probably needs to borrow the pattern used in keyForAttribute (where serialize applies the inverse transformation of deserialize)

PR #9, opened 2016-05-10
Some misc updates

samchrisinger:

### DIFF
--- a/addon/adapters/node.js
+++ b/addon/adapters/node.js
@@ -1,8 +1,14 @@
 import ApplicationAdapter from './application';
 
 export default ApplicationAdapter.extend({
-    buildURL() {
+    buildURL(_, __, ___, requestType) {
 	// Embed contributors
-	return `${this._super(...arguments)}?embed=contributors`;
+	var base = this._super(...arguments);
+	if (requestType === 'GET') {
+	    return `${base}?embed=contributors`;
+	}
+	else {
+	    return base;
+	}
     }
 });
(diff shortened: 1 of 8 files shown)


addon/adapters/node.js
```diff
@@ -1,8 +1,14 @@
 import ApplicationAdapter from './application';
 
 export default ApplicationAdapter.extend({
-    buildURL() {
+    buildURL(_, __, ___, requestType) {
 	// Embed contributors
-	return `${this._super(...arguments)}?embed=contributors`;
+	var base = this._super(...arguments);
+	if (requestType === 'GET') {
```
abought: I was recently looking along parallel lines... does this work? It seems like requestType is the Ember query type, not an HTTP method. See:
https://github.com/emberjs/data/blob/v2.5.1/addon/-private/adapters/build-url-mixin.js#L54

tests/unit/adapters/node-test.js
```diff
@@ -8,6 +8,6 @@ moduleFor('adapter:node', 'Unit | Adapter | node', {
 test('it embeds contributors', function(assert) {
     let adapter = this.subject();
 
-    //let url = adapter.buildURL('nodes', null, null, 'GET', null);
-    assert.ok(adapter);
+    let url = adapter.buildURL('nodes', null, null, 'GET', null);
+    assert.ok(url.indexOf('embed=contributors'));
```
abought: A unit test! 👍

PR #11, opened 2016-05-10
Serializer and adapter cleanup

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-3

# Purpose
Start work on serializing data to send to server

# Notes for Reviewers
- Some "send to server" actions may be incomplete, but posting this for visibility and to avoid rework.
- Feedback on tests is welcome; I think the `extractAttributes` test could be more elegant than it is
- Worth giving a try in the browser to be sure that I got rid of all experiments

This PR had some trial and error; it's a good candidate for "merge and squash"

### Routes Added/Updated
- Add an "edit title" box to node detail page, to test writing to server
- Add a placeholder action for later work on "create node" button

# Blockers/ TODO
- [x] Remove unused uri templates packages
- [x] remove extraneous fields from payload and resolve 500 server errors


addon/mixins/osf-login-route.js
```diff
@@ -10,8 +10,9 @@ export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
         if (config.OSF.isLocal) {
             accessToken = config.OSF.accessToken;
         } else {
-            // Acquire an OSF access token, then exchange it for a Jam token
+            // Acquire an OSF access token, then exchange it for a Jam token  // TODO: Jam?
```
samchrisinger: Shameless copy/paste. Will remove during merge.

index.js
```diff
@@ -5,55 +5,46 @@ var config = require('config');
...
+                ENV.OSF.url = 'https://staging.osf.io/';
+                ENV.OSF.apiUrl = 'https://staging-api.osf.io';
+                ENV.OSF.authUrl = 'https://staging-accounts.osf.io/oauth2/authorize';
+        }
+        if (BACKEND === 'stage2') {
+                ENV.OSF.url = 'https://staging2.osf.io/';
+                ENV.OSF.apiUrl = 'https://staging2-api.osf.io';
+                ENV.OSF.authUrl = 'https://staging2-accounts.osf.io/oauth2/authorize';
+        }
+        if (BACKEND === 'prod') {
```
sloria: Probably shouldn't condone using `prod` as a backend for local development. If a dev needs a prod-like backend, they should use test.osf.io.

samchrisinger: @sloria agreed, and I am trying to get the [README to reflect this](https://github.com/CenterForOpenScience/ember-osf/commit/b631dfe7a57c96612e020541caed18ee3f394efc#diff-04c6e90faac2675aa89e2176d2eec7d8) (we can be more explicit too), but the code should not assume that it's never being run in production.

abought: Adding support for the test backend got lost in the shuffle, but I agree it should be added for use by third-party integrators. Thanks for the reminder!

Presumably, `prod` is kept for use in deployment.

sloria: @samchrisinger As discussed in our other thread, envvars should be used for production and any non-local deploys.

PR #14, opened 2016-05-11
[RFR] Blueprints fixes

samchrisinger:
## Ticket
https://openscience.atlassian.net/browse/EOSF-29

# Purpose

We noticed the ember-osf-settings blueprint was not consistently generating a settings file that was seen by the config module. This change adds a default blueprint that gets generated when the addon is installed. It tries to be clever about merging the config with an existing one.


blueprints/ember-osf/index.js
```diff
@@ -0,0 +1,30 @@
...
+	// h/t: https://github.com/samselikoff/ember-cli-mirage/blob/master/blueprints/ember-cli-mirage/index.js
+        // this prevents an error when the entityName is
+        // not specified (since that doesn't actually matter
+        // to us
+    },
+    afterInstall: function() {
+	var configPath = path.join(this.project.root, 'config', 'local.yml');
+	var tmpConfigPath = path.join(this.project.root, 'config', '_local.yml');
+	try {
+	    debugger;
```
abought: Debugger statement should probably be removed before merging. ;)

index.js
```diff
@@ -1,47 +1,56 @@
 /* jshint node: true */
 'use strict';
+var path = require('path');
 var config = require('config');
 
 module.exports = {
     name: 'ember-osf',
+    blueprintsPath: function() {
+	return path.join(__dirname, 'blueprints');
```
abought: Minor note, but this should be indented per style guide.

(doing it now avoids noisy cleanup commits later)

index.js
```diff
@@ -1,47 +1,56 @@
...
+        if (BACKEND === 'test') {
+            ENV.OSF.url = 'https://test.osf.io/';
+            ENV.OSF.apiUrl = 'https://test-api.osf.io';
+            ENV.OSF.authUrl = 'https://test-accounts.osf.io/oauth2/authorize';
         }
         if (BACKEND === 'prod') {
-                ENV.OSF.url = 'https://osf.io/';
-                ENV.OSF.apiUrl = 'https://api.osf.io';
-                ENV.OSF.authUrl = 'https://accounts.osf.io/oauth2/authorize';
+	    console.log('WARNING: you\'ve specified production as a backend. Please do not use production for testing or development purposes');
```
abought: More indentation, please.

README.md
```diff
@@ -10,37 +10,46 @@ This repo contains code for interacting with the OSF APIv2 inside of an Ember ap
...
+```yaml
+<backend>:
+  CLIENT_ID: null
+  PERSONAL_ACCESS_TOKEN: null
+  OAUTH_SCOPES: osf.full_read osf.full_write
+```
+
+You will need to fill out options for each backend you want to use (see 'Running' below).
+We reccomend using the 'test' backend for development and testing as it is the most stable
+of our staging enviorments.
```
abought: Language notes:

- Spell check
  - Typo: should be `recommend`
  - Typo: should be `environments`

Also, we shouldn't refer to `test` as a staging environment. As far as I know, the original intent was to always mirror master- it's a tool for external integrators, more akin to `demo.dataverse.org` than to a staging environment.

README.md
```diff
@@ -54,27 +63,34 @@ personal access token on your local OSF instance ([here](http://localhost:5000/s
 
 #### Create a local settings file
 
-To do this:
+If for some reason you don't have a config/local.yml you can generate one. To do this:
 ```bash
-ember g ember-osf-settings `echo $HOSTNAME`
+ember g ember-osf `echo $HOSTNAME`
```
abought: Very minor future nitpick: I really like the shorthand commands like `ember g` myself... but when we are explaining this to people not familiar with ember, we should spell out the entire command explicitly. It's a small bit of polish that helps make a new tool more approachable.

`ember generate ...` vs `ember g`, `ember server` vs `ember s`, etc.

The full names are the ones most commonly explained in documentation, and aliases are one more layer that people have to learn about before they can get started.

README.md
```diff
@@ -54,27 +63,34 @@ personal access token on your local OSF instance ([here](http://localhost:5000/s
...
+```javascript
+this.store.findAll('node')
+```
+to fetch all nodes.
 
 ## Running
 
-First, decide which backend you would like to target. Typically we reccomend developers use either our staging or test servers:
-- staging (`stage`): contains bleeding edge features, but less stable
+We reccomend developers target out test server:
```
abought: Typo: `recommend`

blueprints/ember-osf/index.js
```diff
@@ -0,0 +1,30 @@
...
+	    var contents = fs.readFileSync(configPath).toString();
+	    if (!CONFIG_MARKER.test(contents)) {
+		var tmpConfig = fs.readFileSync(tmpConfigPath).toString();
+		fs.appendFileSync(configPath, tmpConfig);
```
abought: Style guide/ indentation.

blueprints/ember-osf/index.js
```diff
@@ -0,0 +1,30 @@
...
+	    }
+	    fs.unlinkSync(tmpConfigPath);
+	} catch (e) {
+	    fs.renameSync(tmpConfigPath, configPath);
```
abought: Cosmetic but confusing: When I run the script, it says it created a file in `_local.yml` (which is nice).

But this script then renames it, so the file is actually in `local.yml`. This may be confusing to new users.

samchrisinger: This is very hard to change (would require overriding ember-cli internals).

README.md
abought: Now that we generate a file with a fixed filename, what is the purpose of the `HOSTNAME` argument?

abought:
I've run the code and it works- excellent!

When I specified an undefined backend, I got an error message which probably needs to be updated:

`WARNING: you have specified a backend 'local' that you have not configured in your config/<hostname>.yml`

blueprints/ember-osf/index.js
abought: What if you changed which one was the temp file? (eg by renaming the existing file before copying over, so that the user is correctly instructed to look at the resulting config file that matters to them) 

As an added advantage, backing up the original file before overwriting it would provide protection if something went wrong. Just overwriting the old file may lead to issues; we may be trying to be *too* smart here.

abought:
Pass complete. It works, and looks like an improvement over what we have in many ways- thanks!  Most of the notes are cosmetic in nature. 

Backing up (rather than overwriting) a pre-existing config may also be a good idea.

blueprints/ember-osf/index.js
samchrisinger: I don't feel very good about renaming a user's config files. What if this crashed before finishing?

> before overwriting it 

It's an append

abought:
Language changes look good.

The file renaming isn't a big enough deal to justify waiting, esp in light of #15.

PR #19, opened 2016-05-11
Feature: File models

aaxelb:
## Ticket
https://openscience.atlassian.net/browse/EOSF-6

# Purpose
Implement file model, add file detail route and a basic file tree to the dummy app.

# Notes for Reviewers
Mirroring an oddity in the api, `node.files` is actually a list of `file-provider`s, and `fileProvider.files` contains actual `file`s. 

### Routes Added/Updated
`file/:file_id`


tests/dummy/app/templates/components/file-tree.hbs
```diff
@@ -0,0 +1,10 @@
+<div style="padding-left:1em">
```
samchrisinger: @aaxelb please don't get carried away with this implementation. A more full-featured file tree will happen: https://openscience.atlassian.net/browse/EOSF-16

samchrisinger: Also please keep css out of handlebars templates, instead use a style.css in this folder https://github.com/CenterForOpenScience/ember-osf/blob/develop/.github/CONTRIBUTING.md#styling-components

aaxelb: Makes sense, thanks

addon/adapters/file-provider.js
```diff
@@ -0,0 +1,7 @@
+import ApplicationAdapter from './application';
+
+export default ApplicationAdapter.extend({
```
samchrisinger: Is this adapter needed? The superclass impl. should already return 'files': https://github.com/emberjs/data/blob/v2.5.2/addon/adapters/json-api.js#L110

aaxelb: Won't it return 'fileProviders'? The /files endpoint returns objects that should be treated as `file-provider`s, is the problem.

samchrisinger: Oh, good call. Thanks for the clarification.

PR #21, opened 2016-05-12
[WIP] Move OSF-specific behavior to separate base classes

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-33

# Purpose
Isolate apiv2-specific behaviors, rather than putting them in the application base classes.

By not making the application adapter/serializer/ base model specific to one API, we have room to support other microservices such as Waterbutler, whose APIs are not JSONAPI compliant. 

This change uses mixins where we just want to add a bundle of properties, and classes where we need a whole different parent (eg not JSONAPI-*).

Utility depends on how we handle files API, of course.

# Notes for Reviewers
Every other concurrent PR dealing with models would be affected by this change, if the new models don't subclass the `OsfAdapter` / `OsfSerializer` (or extend with the `OsfModel` mixin) as appropriate.

## TODO
- [x] Get unit tests passing (it seems that application adapter had special rules and was exempted from dealing with some fiddly bits)
- [x] Update with newest models

### Routes Added/Updated
No routes were added. However, any bugs in data fetching would affect all existing pages.


addon/models/user.js
```diff
@@ -1,8 +1,8 @@
 import DS from 'ember-data';
 
-import OsfModel from 'ember-osf/models/base';
+import OsfModel from '../mixins/osf-model';
```
samchrisinger: Can you explain the motivation for switching to a relative import?

abought: No strong motivation; it seemed that some other files were starting to move this way and I was just matching styles.

It's entirely possible that there's a subtlety of addons where absolute imports would make more sense. If that's the case, I'm more than happy to change it back.

samchrisinger: I think they're equivalent ways of importing the file, but for whatever reason '../' in an import statement has always made me a bit anxious. That said many of the ember blueprints seem to do this...

No change needed, but we should try to be consistent with this going forward.

PR #24, opened 2016-05-13
Feature/file management [EOSF-13]

aaxelb:
## Ticket
https://openscience.atlassian.net/browse/EOSF-13
# Purpose
Allow for uploading/managing files.
# Notes for Reviewers
This provides a `file-manager` service that acts as an interface for waterbutler file actions.

The dummy site now allows browsing a node's files and doing stuff to the files and folders:
* Files: rename, delete, download, update contents, move, copy
* Folders: rename, delete, add subfolder, upload file, move, copy

### Routes Added/Updated
`nodes.detail.files`
`nodes.detail.files.provider`
`nodes.detail.files.provider.file`

Removed `file` route.


addon/services/file-manager.js
```diff
@@ -0,0 +1,114 @@
...
+        var sessionData = this.get('session').get('data');
+        var accessToken = sessionData.authenticated.attributes.accessToken;
+        return new Ember.RSVP.Promise(function(resolve, reject) {
+            Ember.$.ajax(url, {
+                method,
+                headers: {
+                    Authorization: `Bearer ${accessToken}`
+                },
+                error(_, __, errorMsg) {
+                    reject(new Error(errorMsg));
```
samchrisinger: Might be good to return a DS.Errors instance?
http://emberjs.com/api/data/classes/DS.Errors.html

thoughts?

aaxelb: Yeah, definitely, that makes sense.

PR #26, opened 2016-05-16
[Hold] Feature/Node Update Behaviors [EOSF-12]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-12

# Purpose

- Implements more node updates: change description, change category, change privacy
- Add a component - thanks @hmoco for helping me with this!
- Delete node (currently JS error if try to delete node with children)

### Routes Added/Updated
Added /nodes/:node_id/children/


addon/serializers/children.js
```diff
@@ -0,0 +1,12 @@
+import OsfSerializer from './osf-serializer';
+
+export default OsfSerializer.extend({
+    serialize: function(snapshot, options) {
```
abought: Perhaps `payloadKeyFromModelName` is what you're looking for to set the type for an individual item?
http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_payloadKeyFromModelName

Also, try not to duplicate code defined in the parent (eg deleting the relationships section). You can use `this._super(...arguments)` to do things already defined in the parent method.

addon/adapters/children.js
```diff
@@ -0,0 +1,11 @@
+import OsfAdapter from './osf-adapter';
+
+export default OsfAdapter.extend({
+    buildURL(_, __, snap, requestType) {
+        // Embed contributors
```
abought: How does this adapter embed contributors? It's not clear they're mentioned from the code below; a more descriptive comment might be helpful for some of the trickery below.

addon/serializers/children.js
pattisdr: From http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_payloadKeyFromModelName
> You can use payloadKeyFromModelName to override the root key for an outgoing request.

I don't want to override the root key, I want to change the type from 'children' to 'nodes', because we don't support the 'children' type in the API.

abought: Like this?
https://github.com/emberjs/data/blob/v2.5.1/addon/serializers/json-api.js#L397

pattisdr: Yes, but `snapshot.modelName` is `children`.  I don't want to use the modelName, I want to override it.

pattisdr: Okay, after talking with @abought, he's right, I just need to upload the payloadKeyFromModelName hook on the serializer, which sets the type.

samchrisinger:
Closing in favor of https://github.com/CenterForOpenScience/ember-osf/pull/47

PR #23, opened 2016-05-13
[WIP] Merging embeds + mirage

samchrisinger:
Replacing https://github.com/CenterForOpenScience/ember-osf/pull/12

// TODO

PR #29, opened 2016-05-16
[Ready] Feature/node contrib management [EOSF-38]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-38

# Purpose
- [x] Add contributor
- [x] Update contributors 
- [x] Remove contributor

# Notes for Reviewers

Bulk update functionality https://openscience.atlassian.net/browse/EOSF-42 will improve this feature.  Right now, you can update multiple contributors in the dummy app, but it is making single, consecutive requests in the background to update each contributor.  There's no guarantee currently that these requests happen in the correct order.     This can cause problems, if for example, there are two contributors on the project, and I am the only admin. if I attempt to make you the admin, and change my permissions to read, this has to happen in the correct order.  You need to be an admin before I can remove myself as admin.

If bulk update functionality isn't eventually implemented, there could be additional checks to ensure the requests happen in the correct order.

### Routes Added/Updated

None.

### DIFF
--- a/tests/dummy/app/routes/nodes/detail.js
+++ b/tests/dummy/app/routes/nodes/detail.js
@@ -64,13 +64,22 @@ export default Ember.Route.extend({
             }
         }
     },
-    canRemoveContributor(contribRemoving, contribMap) {
-        /** Contributor can only be removed if there is at least one other contributor
-        with admin permissions, and at least one other bibliographic contributor **/
+
+    generateContributorMap(contributors) {
+        // Maps all node contributors to format {contribID: {permission: "read|write|admin", bibliographic: "true|false"}}
+        var contribMap = contributors.content.currentState.reduce(function(newMap, contrib) {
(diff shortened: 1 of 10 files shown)


tests/dummy/app/routes/nodes/detail.js
```diff
@@ -21,7 +22,141 @@ export default Ember.Route.extend({
             } else {
                 console.log('You do not have permissions to edit this node');
             }
+        },
+        addContributor(contribId, permission, bibliographic) {
+            var node = this.modelFor(this.routeName);
+            if (contribId) {
+                if (node.get('currentUserPermissions').indexOf('admin') !== -1) {
```
samchrisinger: Adding a TODO for myself to update permissions strings when https://github.com/CenterForOpenScience/ember-osf/pull/23/files#diff-7fd0bf247bef3c257e0fcfd7e544a338R5 is merged.

PR #33, opened 2016-05-18
Add file version component [ESOF-18]

samanehsan:
## Ticket
https://openscience.atlassian.net/browse/EOSF-18

# Purpose
Add a file version component

# Notes for Reviewers
None

### Routes Added/Updated
None


samchrisinger:
Looks great! Can you please generate this as a component outside of the dummy app? Just
`ember g component file-version`
and move the relevant code to `addon/components`?

samanehsan:
@samchrisinger, what code should go in `addon/components`?

samanehsan:
Also, should I still use the pod structure for components?

samchrisinger:
@samanehsan yes please use pod structure for all components. 

The way this feature is implemented it is only visible/accessible inside of the dummy app. Since we're developing an [Ember addon](https://gist.github.com/kristianmandrup/ae3174217f68a6a51ed5) we'll need to put the code we expect consuming Ember apps to use inside of the addon and app folders.

Your file structure should eventually look like:
- ember-osf/
  - addon/
    - components/
      - file-version/
         - component.js, template.hbs, style.css

and

- ember-osf/
  - app/
    - components/
      - file-version/
         - component.js

I recommend using the ember-cli to generate the new files first, then moving the old code over to the correct places.

samanehsan:
Got it! Thanks! 🐙

samanehsan:
Done making that change @samchrisinger! And the build is passing now :octocat:

PR #36, opened 2016-05-19
Generic resource pagination for routes

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-30

# Purpose
Allow routes to work with paginated queries: fetch more than just the first 10 records that a bare API call would normally return.

This provides a pair of mixins (route and controller) that add pagination query params to a route
(`page` and `page_size`- these names will be invisibly translated to those used by the API backend). These mixins add next, previous, and goto actions, as well as basic total page tracking. Because API pagination depends on query params, there is also a helper function to construct the appropriate URL query.

The serializer has also been updated to capture basic pagination metadata for routes that return multiple results. (though actual use of pagination must always go through a query, since the API relies on url params)

screen shot 2016-05-19 at 6 56 22 pm

# Notes for Reviewers
This supports pure pagination by re-firing the model hook when url params change. It does not support use cases such as infinite scrolling.

Comments on the API would be welcome- eg, having to pass model params in every time is kind of awkward.

Since the new HTML linter insists on 2 spaces, I added a small erratum to make `.editorconfig` respect that setting.

# TODO
- [x] Research implementations
- [x] Add route and controller mixins
- [x] Retrofit a route to demonstrate pagination in action
- [x] Make pagination metadata (next/prev/self links) accessible to model
- [x] Add sample components for bootstrap-style pagination controls
- [x] Make pagination url args configurable to support non-apiv2 resources

### Routes Added/Updated
The users page has been updated with primitive next/back buttons, to demonstrate how these mixins can be used.


tests/dummy/app/routes/users/list.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+
+import PaginatedRouteMixin from  'ember-osf/mixins/paginated-route';
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, PaginatedRouteMixin, {
+    store: Ember.inject.service(),
```
abought: Note to self: in a separate PR, go through routes and clean up any place where `store` is being injected manually into the route. (this boilerplate is unnecessary)

Ditto for `session` if the route doesn't explicitly use it.

PR #37, opened 2016-05-19
Feature/MFR File Renderer [EOSF-19]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-19

Dependent on https://github.com/CenterForOpenScience/modular-file-renderer/pull/215

# Purpose
Create an Ember component to render a file.

TODO
- [x] Create file-renderer component
- [x] Add render urls to config
- [x] Figure out how to build MFR url
- [x] PR to MFR so it can handle waterbutler v1 urls

# Notes for Reviewers

### Routes Added/Updated

Stuck the file rendering on the file detail page in the dummy app.


addon/components/file-renderer/template.hbs
```diff
@@ -0,0 +1,2 @@
+<iframe src={{buildMfrUrl download}} width="100%" height=100% scrolling="no" marginheight="0" frameborder="0">
```
samchrisinger: @pattisdr do we have a use case for building MFR urls outside of this component? Using a template helper seems like unneeded indirection. Since the url is a function of the `download` property of the component I'd rather see it moved to a computed value there. 

E.g.
```javascript
   layout,
   download: null,
   mfrUrl: Ember.computed('download', function() {
       ...
   })
});
```

And you template can just:
```html
<iframe src={{mfrUrl}} ... 
```

samchrisinger: Also ember lets you [pass DOM attributes directly to components](https://guides.emberjs.com/v2.4.0/components/customizing-a-components-element/#toc_customizing-the-element). In this case `width="100%" height=100% ` don't seem like things we should rigidly impose on consumers. A better api could be:

```html
{{file-renderer download=download width="100%" height="100%"}}
```

You may need to [set the component's `tagName: ''`](http://emberjs.com/api/classes/Ember.Component.html#property_tagName). By default Ember wraps every component in a div. Not 100% were the passed attributes will end

addon/components/file-renderer/template.hbs
```diff
@@ -0,0 +1,2 @@
+<iframe src={{mfrUrl}} width={{width}} height={{height}} scrolling="no" marginheight="0" frameborder="0">
```
samchrisinger: @pattisdr might be a good idea to provide reasonable defaults in the component.js,
i.e. 
```
height: '100%',
width: '100%'
```

samchrisinger: I'm going to take care of this during merge.

PR #39, opened 2016-05-24
Add ember-data factories for tests

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-39

# Purpose
Add factories to create models for unit tests.

Preloads a few sample values for choice fields and uses `faker` library to generate sample data. Relationships data super-untested.

# Notes for Reviewers

- Some aspects- especially relationships generation- are as yet untested in an applied case.
- Can use multiple traits to add optional sections to a record.
- [ ] Why do we pin to an old version of faker (3.0 instead of 3.1)? The newer has some facilities for, eg, filename generation.

This PR depends on phantomJS > 2.0 versions, and adds a dependency. Advice on how to make travis more efficient welcome.

### Routes Added/Updated
None.

Updated the integration tests for the file-version component, to show usage of factory. (incl required setup steps)


abought:
CI is now passing, thanks to a new `phantomjs-prebuilt` dependency.

Future improvement: consider using cached directories for node modules. (apparently some people were seeing intermittent failures due to rate limiting- this is a popular download for CI powerusers)

PR #40, opened 2016-05-24
Add Component and Relationships

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-36
https://openscience.atlassian.net/browse/EOSF-12 (in part)

# Purpose
Implement relationship endpoint logic, and custom serializers for relationship attributes. Includes dummy app use cases of adding and removing institution affiliation and adding a component to a node.

### Routes Added/Updated
node/details updated to include the buttons to tests relationship functionality.


addon/adapters/osf-adapter.js
```diff
@@ -11,13 +12,55 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+        if (relationMeta.options.serializer) {
+          serialized = relationMeta.options.serializer(snapshot);
+        } else {
+          var serializer = store.serializerFor(relationType);
+          var toBeSent = snapshot.record.get(dirty).filter(
+            function(record){
+              return record.id === null;
+            }
+          );
+          //under the assumption relationship saves are atomic, only one component is added at a time.
```
samchrisinger: This has a different implication. We want the assumption that only one related field can be changed at a time not that only one record per field can change at a time.

For the institutions case, for example, we to update the Node with all of the added institutions. Once https://openscience.atlassian.net/browse/EOSF-42, this behavior will also work for contributors/components.

hmoco:
@samchrisinger, I think this is ready for a look.

addon/mixins/osf-model.js
```diff
@@ -7,5 +7,37 @@ import DS from 'ember-data';
...
+            if (dirtyRelationships[rel]) {
+                return rel;
+            }
+            return null;
+        }).filter(Boolean);
+    }),
+    _dirtyRelationships: {},
+    onLoad: Ember.on('ready', function() {
+        this.eachRelationship((rel, meta) => {
+            var _this = this;
```
samchrisinger: why `_this`?

samchrisinger:
```javascript
var update = () => this.set(`_dirtyRelationships.${rel}`, !Ember.isEmpty(_this.get(`_dirtyRelationships.${rel}`)));
```

addon/models/node.js
```diff
@@ -2,6 +2,16 @@ import DS from 'ember-data';
 
 import OsfModel from '../mixins/osf-model';
 
+function relationshipSerializer(relationship, type) {
```
samchrisinger: This should go in a non node-specific util file

addon/mixins/osf-model.js
hmoco: Test suite wasnt happy with ` bind `, and linter complains that ` self ` should be ` _this `

abought: The `bind` issue for tests should be resolved by upgrading to PhantomJS >= 2.0; see flowdock discussion yesterday. We ran into this on Travis CI.

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,54 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+        ) : false;
+        if (links) {
+            url = links.self ? links.self.href : links.related.href;
+        } else {
+            url = this._super(...arguments);
+        }
         // Fix issue where CORS request failed on 301s: Ember does not seem to append trailing
         //  slash to URLs for single documents, but DRF redirects to force a trailing slash
-        if (url.lastIndexOf('/') !== 0) {
+        if (url.lastIndexOf('/') !== url.length - 1) {
```
abought: May want to change this from `lastIndexOf` to to `indexOf` to preserve intended meaning.

(checking for a trailing slash)

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,54 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
     authorizer: 'authorizer:osf-token',
     host: config.OSF.apiUrl,
     namespace: config.OSF.apiNamespace,
-
-    buildURL(modelName, id, snapshot, requestType, query) {  // jshint ignore:line
-        var url = this._super(...arguments);
+    buildURL(modelName, id, snapshot, requestType, query, dirtyRelationship) { // jshint ignore:line
```
abought: Could you please add a source code comment explaining the design intent for this section?

(to help future maintainers as they work out what each of the things in `buildURL` is intended to do)

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,54 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+        }
         // Fix issue where CORS request failed on 301s: Ember does not seem to append trailing
         //  slash to URLs for single documents, but DRF redirects to force a trailing slash
-        if (url.lastIndexOf('/') !== 0) {
+        if (url.lastIndexOf('/') !== url.length - 1) {
             url += '/';
         }
         return url;
+    },
+    relationshipPayload(snapshot, dirty, store) {
```
abought: This looks a lot like serialization logic- can you explain why it is defined in the adapter?

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,54 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+            serialized = serializer.serialize(new DS.Snapshot(toBeSent[0]._internalModel));
+            // for some reason this is not hitting the node overloaded serialize method
+            delete serialized.data.relationships;
+        }
+        return serialized;
+    },
+    updateRecord(store, type, snapshot, _, query) {
+        var dirtyRelationships = snapshot.record.get('dirtyRelationships');
+        if (dirtyRelationships.length) {
+            var dirty = dirtyRelationships.pop();
```
abought: `pop` returns the last element from the array. What happens if more than one relationship is dirty? Is that handled / does it need to be?

addon/mixins/osf-model.js
```diff
@@ -7,5 +7,37 @@ import DS from 'ember-data';
 
 export default Ember.Mixin.create({
     links: DS.attr('links'),
-    embeds: DS.attr('embed')
+    embeds: DS.attr('embed'),
+
+    relationshipLinks: DS.attr('links'),
```
abought: This seems to be the same content as appears in `links`- prefer one clear field of choice for a piece of data when possible.

addon/mixins/osf-model.js
```diff
@@ -7,5 +7,37 @@ import DS from 'ember-data';
...
+    dirtyRelationships: Ember.computed('_dirtyRelationships', function() {
+        var dirtyRelationships = this.get('_dirtyRelationships');
+        return Object.keys(dirtyRelationships).map((rel) => {
+            if (rel === 'files' || rel === 'node' || rel === 'user') {
```
abought: I'm not sure that I'm comfortable hardcoding individual model-specific behavior (node, files, etc) into the overall parent model- this seems like a code smell. 

Why are some models special-cased and others not?

addon/mixins/osf-model.js
```diff
@@ -7,5 +7,37 @@ import DS from 'ember-data';
...
+                return null;
+            }
+            if (dirtyRelationships[rel]) {
+                return rel;
+            }
+            return null;
+        }).filter(Boolean);
+    }),
+    _dirtyRelationships: {},
+    onLoad: Ember.on('ready', function() {
```
abought: Not a big deal, but instead of adding an observer, can you just add a function named `didLoad` to use existing events?  Eg
http://jsfiddle.net/pangratz666/Vx8yL/

(Honestly asking. But if we can use an existing method hook, that's preferable)

addon/mixins/osf-model.js
abought: With the newer phantomJS version, you should be able to simplify this back to `addObserver(update.bind(this))` or similar- please do.

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,54 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+            var url = this.buildURL(type.modelName, snapshot.id, snapshot, 'updateRecord', query, dirty);
+            var requestType = snapshot.record[dirty].meta().options.updateRequestType;
+            return this.ajax(url, requestType || 'PATCH', {
+                data: this.relationshipPayload(snapshot, dirty, store)
+            });
+        } else {
```
abought: If I'm reading this correctly, `updateRecord` can only save relationships OR payload fields, but not both at once. (as determined by the `if` statement) Is that the intent?

addon/adapters/osf-adapter.js
abought: It seems that we're forcing the same save method to handle two separate kinds of requests; I think I'd almost prefer a separate `buildURL` method for relationships. (especially since the relationship functionality doesn't seem to require access to the parent method via `this._super`)

The other advantage of doing things that way is we wouldn't need to extend/alter the call signature of the function. (which could be fragile if ember added their own arguments in the future)

addon/models/node.js
```diff
@@ -25,10 +35,12 @@ export default DS.Model.extend(OsfModel, {
...
     }),
     children: DS.hasMany('nodes', {
-        inverse: 'parent'
+        inverse: 'parent',
+        updateRequestType: 'POST'
     }),
     affiliatedInstitutions: DS.hasMany('institutions', {
-        inverse: 'nodes'
+        inverse: 'nodes',
+        serializer: relationshipSerializer('affiliatedInstitutions', 'institutions')
```
abought: Why do institutions use a custom serializer, but not children? Can we add any documentation to clarify the very special, complicated rules in play for these relationships?

tests/dummy/app/routes/nodes/detail.js
```diff
@@ -3,28 +3,47 @@ import Ember from 'ember';
...
+        controller.set('editedTitle', model.get('dscription'));
         this._super(...arguments);
     },
 
     actions: {
-        editExisting(value) {
+        editExisting(title, description, category, isPublic) {
             // TODO: Should test PUT or PATCH
-            console.log('Will edit title from', this.modelFor(this.routeName).get('title'), ' to ', value);
+            // console.log('Will edit title from', this.modelFor(this.routeName).get('title'), ' to ', value);
```
abought: Instead of commenting out code, we should remove it when unnecessary.

tests/dummy/app/routes/nodes/detail.js
```diff
@@ -3,28 +3,47 @@ import Ember from 'ember';
...
+                if (category) {node.set('category', category);}
+                if (description) {node.set('description', description);}
+                if (isPublic !== null) {node.set('public', isPublic);}
                 node.save();
             } else {
                 console.log('You do not have permissions to edit this node');
             }
         },
+        affiliateNode(instId) {
+            var _this = this;
```
abought: If you use an ES6 arrow function in the promise, then `var _this = this;` should be unnecessary, and should be removed.

addon/adapters/osf-adapter.js
hmoco: yes!

addon/models/node.js
hmoco: Because children can use the node serializer, but institutions (and other relationships that use a `/relationships/model` ) require a specific (simpler) serialization of the relationship contents.

addon/adapters/osf-adapter.js
hmoco: we are assuming, at the least for this PR, that .save() are atomic, changing one relationship or attributes at any one time (can change multiple attributes or a single relationship).

addon/mixins/osf-model.js
hmoco: @samchrisinger

addon/mixins/osf-model.js
hmoco: That was because our dirty-ing method was catching those fields as well, and they are not fields that would be updated via relationships anyways (we have children and contributor for that)

addon/adapters/osf-adapter.js
abought: Thanks. Perhaps we can throw an error to enforce that assumption? 

If we don't track these major assumptions, this could lead to very subtle bugs in the future.

tests/dummy/app/templates/nodes/detail/index.hbs
```diff
@@ -2,18 +2,79 @@
...
+                <td>{{input value=editedTitle}}</td>
+            </tr>
+            <tr>
+                <td><label> Description: </label></td>
+                <td>{{input value=editedDescription}} </td>
+            </tr>
+        </table>
+        <p><label> Category: </label></p>
+        <div>
+            <p>{{radio-button name="category" value="analysis" checked=editedCategory}} Analysis </p>
```
abought: Not in scope of this PR, but in the future, it would be nice to break this category selector into a separate component.  I could see this being useful and reusable. 

(and maybe change to a dropdown box)

tests/dummy/app/templates/nodes/detail/index.hbs
```diff
@@ -2,18 +2,79 @@
...
+            <p>{{radio-button name="category" value="instrumentation" checked=editedCategory}} Instrumentation </p>
+            <p>{{radio-button name="category" value="methods and measures" checked=editedCategory}} Methods and Measures </p>
+            <p>{{radio-button name="category" value="procedure" checked=editedCategory}} Procedure </p>
+            <p>{{radio-button name="category" value="project" checked=editedCategory}} Project </p>
+            <p>{{radio-button name="category" value="software" checked=editedCategory}} Software</p>
+            <p>{{radio-button name="category" value="other" checked=editedCategory}} Other </p>
+            <p>{{radio-button name="category" value="uncategorized" checked=editedCategory}} Uncategorized </p>
+        </div>
+        <p><label> Privacy settings </label></p>
+        {{radio-button name="isPublic" value=1 checked=isPublic}} Public
```
abought: A checkbox would make more sense here. (there are only two options)

abought:
Pass complete.

There are a few minor elements of cleanup / duplication / simplifying that would be quick to add (and worth doing). Also a request to separate concerns. (eg maybe avoid having one buildUrl method for two distinct kinds of operation)

Architecturally speaking, the biggest point of concern is that there are several ways in which these changes could lead to unexpected partial saves. (we'd be relying on implementers to know about, and account for, a model behavior that worked differently from normal ember)

In these cases, I strongly recommend that we look for a good way to either:
a) Throw errors if data is not correctly saved, or,
b) Allow all data to be saved in one operation as desired, eg via helper methods that chain saves

Just so long as we don't leave things partially saved without being clear.

This would be a "sooner rather than later" situation, as we don't want to forget about it too long- it could create subtle bugs for users otherwise.

abought:
Overall the approach seems to make sense, and there's a lot less ember witchcraft than I'd expected- yay there! 

Given that we're adding a lot of extra and specialized behavior, I'd definitely encourage us to add unit tests. That could reasonably be done in a separate PR, so as to not block people waiting for the basic functionality. 

Since relationships are a fairly core feature, I'd propose that the unit testing be a pre-1.0 blocker, though.

addon/mixins/osf-model.js
abought: Right- but I'm not really comfortable baking the list of specific exceptions into a parent class. 

a) Parent classes aren't supposed to be model-specific, and b) I could imagine a future model with a relationship that we didn't want to exempt.

If we had a relationship mixin, it could define a simple field at the top level:
`relationshipsNotToTrackInFancyMagic = ['node']` 

...and that single field could be overridden/checked per subclass as appropriate. That would provide a simple hook to control this behavior in a model-specific way.

(to the implementer, it would look kind of like `Component.classNames`)

addon/adapters/osf-adapter.js
abought: Er, that'd also be weird. See this page for a suggested endsWith polyfill:
http://stackoverflow.com/questions/280634/endswith-in-javascript

...I originally tried using ES6 `endsWith`, but encountered errors because it was supported in chrome, but not phantomJS. As long as the intent is preserved, I'm happy with whatever works.

samchrisinger: @abought `lastIndexOf` returns a positively indexed position. So I think this change is appropriate.

PR #44, opened 2016-05-25
ARIA + Internationalization demo

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-25

# Purpose
Apply a few best practices for translation and ARIA labeling in ember-osf. Use the existing signup page as a test case.

## "Before" pictures
Forms:
screen shot 2016-05-31 at 5 38 08 pm

Landmark regions:
screen shot 2016-05-31 at 5 38 58 pm

WAVE tool results:
screen shot 2016-05-31 at 6 22 44 pm

## "After" pictures
Forms:
screen shot 2016-05-31 at 5 26 30 pm

Landmark regions:
screen shot 2016-05-31 at 5 25 13 pm

WAVE tool results:
screen shot 2016-05-31 at 6 23 33 pm

# Notes for Reviewers
Used VoiceOver and WAVE toolbar for the accessibility audit. I can walk you through to see what this looks like in practice.

TODO:
- [x] ARIA best practices for forms
- [x] Unique element ids
- [x] Keyboard focus improvements
- [x] Get it working with both addons and application (deal with namespacing issues)- see [guide](https://github.com/jamesarosen/ember-i18n/issues/255)
- [x] Document required config for consumer app

## Routes Added/Updated
Signup and login components can both be seen on this page: http://localhost:4200/signup

# Proposed guidelines
## Internationalization
- For the addon, put translated text in the files `addon/locales/en-us/translations.js`. [See docs](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Defining-Translations) for translation guide.
- Name translation strings based on the role in the page, not by the text you are replacing. If the element with the text has an ID, name the translation string after the ID.
  - In the component `signup`, the element `signup-heading-title` might have a translation string called `signup.headingTitle`
  - All heading strings for a component should be namespaced based on the component. Component `login-form` would keep all its strings in a translation section called `loginForm`.
- All addon translations should be namespaced under the addon name, to avoid colliding with any translation settings from the app or other addons. 
  - We use `eosf` for this addon

## Accessibility
- Components are reusable, and may therefore add two elements with the same ID to the page. (this can confuse screen readers for, eg, forms). 
  - All IDs should be given unique names by prefixing the component ID to the description, via the elem-id helper, following the syntax `{{elem-id this 'componentnameprefix-roleinpage-descriptor'}}`
- Markup should be used to communicate structure of the document, not just for layout
  - If you want to style a small heading, use CSS. Don't skip from heading level 1 to heading level 3; screen readers use headings to build an outline of the page.
  - Top-level components may want to wrap in an `article` tag to identify themselves as discrete regions of the page. Using `role="region"` (and associating with an aria-label) helps the screenreader jump to that area of the page (eg voiceover "landmarks" menu)
- For forms: 
  - Associate all form fields with labels. 
  - Use HTML5 validation attributes (like `required`) and custom types (like `email`) where relevant
- Use [ARIA roles](https://www.w3.org/WAI/PF/aria/roles#contentinfo) to assign meaning to areas of the page (like warning message alerts)

## Testing
- Do not test based on the presence of english-language words- someone else might have different locale/translation settings. (and adding ember-i18n to tests is an annoying chore)
- Element ids now have a unique prefix; select them via the `id$=` selector ("id ends with")
- If your test does need to check for language, see the [Ember i18n docs- pointers for writing tests](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Testing) for info on how to test this

## Reading list
- Basic checklist: http://a11yproject.com/checklist.html
- Ember Accessibility day / reading list: http://emberjs.com/blog/2016/05/17/global-accessibility-awareness-day.html
- VoiceOver guide: http://accessibility.psu.edu/screenreaders/voiceover/
- WAVE accessibility evaluation Chrome addon: https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh?hl=en-US
- MDN accessibility portal https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA
- Ember i18n docs- pointers for writing tests https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Testing


aaxelb:
I've gone through most of the reading list and am getting started on reviewing the code. I like your proposed guidelines -- thorough and useful.

Out of curiosity, are the guidelines meant for consumers of ember-osf, too, or just for our own development? In other words, can we give a11y for free, or will it require developers using our addon to decide they want it enough to learn about and follow standards too?

addon/instance-initializers/ember-osf.js
```diff
@@ -0,0 +1,16 @@
+/*
+Automatically expose translations for addon in a way that can be merged in with app
+    https://github.com/jamesarosen/ember-i18n/issues/255
+ */
+//import Ember from 'ember';
```
aaxelb: You can remove this commented line

addon/locales/en-us/config.js
```diff
@@ -0,0 +1,16 @@
+// Ember-I18n includes configuration for common locales. Most users
+// can safely delete this file. Use it if you need to override behavior
+// for a locale or define behavior for a locale that Ember-I18n
+// doesn't know about.
+export default {
```
aaxelb: Is this here for future use, or should we delete it?

tests/dummy/app/locales/en-us/config.js
```diff
@@ -0,0 +1,16 @@
+// Ember-I18n includes configuration for common locales. Most users
+// can safely delete this file. Use it if you need to override behavior
+// for a locale or define behavior for a locale that Ember-I18n
+// doesn't know about.
+export default {
```
aaxelb: Same as above, is there a reason to keep this file?

aaxelb:
Looks good to me 👍

addon/locales/en-us/config.js
abought: Technically I think `ember-i18n` requires the file to be present, even if we don't set any options.  (I got errors when I omitted this file)

I debated about leaving the commented out text (from the template) in place, but decided it might be useful sometime; documentation is a smidge thin.

abought:
> Out of curiosity, are the guidelines meant for consumers of ember-osf, too, or just for our own development? In other words, can we give a11y for free, or will it require developers using our addon to decide they want it enough to learn about and follow standards too?

Thanks for giving this a look! 

The short answer: a lot of web accessibility depends on the HTML markup you use. If developers use our premade components, then those components will build in common ARIA features right out of the box. 

But if they write their own code (or pages), they will need to add their own ARIA tags, ensure IDs are unique, etc.

addon/instance-initializers/ember-osf.js
abought: Yes. Yes I should.

Done in commit 13f1ae2 .

PR #47, opened 2016-05-27
Relationships are hard

samchrisinger:
Replacing https://github.com/CenterForOpenScience/ember-osf/pull/40


addon/models/osf-model.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+    links: DS.attr('links'),
+    embeds: DS.attr('embed'),
+
+    relationshipLinks: Ember.computed.alias('links.relationships'),
```
abought: Out of curiosity, what is the benefit of having this alias/ additional field at all?

abought: Durgh. Sorry. Eyes working now.

samchrisinger: Also I want to have a placeholder in place in case we discover a way to not shove relationships in the links field.

tests/dummy/app/routes/nodes/detail.js
```diff
@@ -3,28 +3,47 @@ import Ember from 'ember';
...
+                if (description) {node.set('description', description);}
+                if (isPublic !== null) {node.set('public', isPublic);}
                 node.save();
             } else {
                 console.log('You do not have permissions to edit this node');
             }
         },
+        affiliateNode(instId) {
+            var _this = this;
+            _this.store.findRecord('institution', instId).then(
```
abought: We can eliminate `var _this = this` with an arrow function.

(Reconciling new PR with previous comments)
https://github.com/CenterForOpenScience/ember-osf/pull/40#discussion_r64815604

tests/dummy/app/routes/nodes/detail.js
```diff
@@ -3,28 +3,47 @@ import Ember from 'ember';
...
+        controller.set('editedTitle', model.get('dscription'));
         this._super(...arguments);
     },
 
     actions: {
-        editExisting(value) {
+        editExisting(title, description, category, isPublic) {
             // TODO: Should test PUT or PATCH
-            console.log('Will edit title from', this.modelFor(this.routeName).get('title'), ' to ', value);
+            // console.log('Will edit title from', this.modelFor(this.routeName).get('title'), ' to ', value);
```
abought: Please remove commented out code before merging.

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,82 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+                var requestType = snapshot.record[relationship].meta().options.updateRequestType;
+                return this.ajax(url, requestType || 'PATCH', {
+                    data: this._relationshipPayload(store, snapshot, relationship)
+                }).then(() => snapshot.record.clearDirtyRelationship(relationship));
+            }));
+        }
+        if (Object.keys(snapshot.record.changedAttributes()).length) {
+            promises.push(this._super(...arguments));
+        }
+        return Ember.RSVP.Promise.all(promises).then(values => {
```
abought: Docs on this method aren't great- but it appears that this will be rejected at the first failure. What happens if one save succeeds, but others fail?

This might be relevant and useful:
http://emberjs.com/api/classes/RSVP.html#method_allSettled

tests/unit/utils/serialize-relationship-test.js
```diff
@@ -0,0 +1,7 @@
+// import { serializeHasMany } from 'dummy/utils/serialize-relationship';
```
abought: Is it safe to remove the commented-out code here?

addon/models/osf-model.js
```diff
@@ -0,0 +1,50 @@
...
+        this.set(`_dirtyRelationships.${relationship}`, false);
+        // Also clean the inverse relationship
+        var relatedMeta = this[relationship].meta();
+        // A slight hack to get the related record w/o an API call
+        if (relatedMeta.kind === 'hasMany') {
+            this._internalModel._relationships.initializedRelationships[relationship].canonicalState.forEach(internalModel => {
+                internalModel.record.set(`_dirtyRelationships.${relatedMeta.options.inverse}`, false);
+            });
+        }
+        // else {
```
abought: Can we remove this commented out code?

addon/models/osf-model.js
```diff
@@ -0,0 +1,50 @@
...
+        var dirtyRelationships = this.get('_dirtyRelationships');
+        return Object.keys(dirtyRelationships).filter(k => dirtyRelationships[k]);
+    }).volatile(),
+    clearDirtyRelationship: function(relationship) {
+        this.set(`_dirtyRelationships.${relationship}`, false);
+        // Also clean the inverse relationship
+        var relatedMeta = this[relationship].meta();
+        // A slight hack to get the related record w/o an API call
+        if (relatedMeta.kind === 'hasMany') {
+            this._internalModel._relationships.initializedRelationships[relationship].canonicalState.forEach(internalModel => {
```
abought: Accessing a chain of two private methods makes me nervous, but we'll just have to make sure this is well tested at a near future date.

addon/models/osf-model.js
```diff
@@ -0,0 +1,50 @@
+/*
```
abought: I can understand treating this as a base class, but it seems that ember data compiles a list of available models in the app based on the contents of this directory. 

screen shot 2016-05-27 at 3 22 43 pm


I'm not 100% comfortable with the potential consequences of registering a model we don't intend to be used as such- but it's more a sense of ember magic that could come back to haunt us, rather than a concrete blocking concern.

abought:
During manual testing, Sam and I are both seeing trouble adding a component under a node; he is investigating the relationships code.

abought:
Pass complete for now. Main points:
1. [ ] This PR should pass manual testing before merge. Suggested test cases include adding an institution and adding a child component to a node.
2. [ ] In order to keep the store in sync, we may want to handle mixed successes/failures more elegantly with our chain of promises. See: https://github.com/CenterForOpenScience/ember-osf/pull/47#discussion-diff-64953237

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,82 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+            var updated = values.shift() || {};
+            values.forEach(value => Ember.merge(updated, value));
+            if (Object.keys(updated).length) {
+                return updated;
+            }
+            return null;
```
abought: Normally, it seems that the promise from this method will resolve to the value from the AJAX request. Will changing it to resolve to `null` break semantics for anything that consumes the output of this method with a promise downstream?

tests/unit/utils/serialize-relationship-test.js
samchrisinger: Left in place for when the test gets implemented

addon/models/osf-model.js
samchrisinger: Left in place for when we implement this branch

addon/models/osf-model.js
samchrisinger: agreed

addon/adapters/osf-adapter.js
samchrisinger: good call

addon/adapters/osf-adapter.js
samchrisinger: Ember data tries to update the record in the store based on the return from updateRecord. This is great for last_updated, but totally breaks for relationship requests. Returning null tells ember-data to skip trying to update the record.

abought: Oooh, that's tricky: this method currently covers both the record itself, AND relationship requests. 

Are we turning off too much? Eg could we resolve to just the value from the super call, if applicable?

addon/adapters/osf-adapter.js
```diff
@@ -80,7 +85,7 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
         if (Object.keys(snapshot.record.changedAttributes()).length) {
             promises.push(this._super(...arguments));
         }
-        return Ember.RSVP.Promise.all(promises).then(values => {
+        return Ember.RSVP.Promise.allSettled(promises).then(values => {
```
abought: Not every value returned from this will be successful...

If I'm reading this correctly, then to use this method and update the store correctly we'll need to do something (like filtering?) to check based on the `state` parameter of the return object. See linked docs for info.

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,82 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+    },
+    /**
+     * Construct a URL for a relationship create/update/delete. Has the same
+     * signature as buildURL, with the addition of a 'relationship' param
+     *
+     * @method _buildRelationshipURL
+     * @param {String} relationship the relationship to build a url for
+     * @return {String} a URL
+     **/
+    _buildRelationshipURL() {
```
abought: Specifically because this is the definition of the function, we should try to specify the parameters in the call signature- especially since we control the method, and all usages. Can we please avoid using destructuring this way?

addon/adapters/osf-adapter.js
samchrisinger: Possibly, but I think this has the same effect. Seems like extra logic for little benefit.

abought:
Basic creation functionality works- yay!

However, it appears the state of the store isn't being updated quite right. I'm not sure whether that's due to `allSettled` semantics we discussed, or something else.

- When I affiliate a node with the `cos` institution ID on staging, the list of affiliated institutions remains blank (and the textbox value doesn't clear)
screen shot 2016-05-27 at 4 45 27 pm

- When I refresh the page, the institution affiliation shows up correctly.
screen shot 2016-05-27 at 4 44 50 pm

It may be that just getting the capability in at all is sufficient victory to merge this i, and ticket further improvements as a followup- Saman also observed some oddness last week in dealing with an updated record.

pattisdr:
I still can't create a component - getting `osf-adapter.js:91 Uncaught TypeError: _ember.default.RSVP.Promise.allSettled is not a function`.

samchrisinger:
@abought @pattisdr I think the promise resolution in updateRecord is all fixed. I'm able to create components against the `test-api`. Staging is dead atm so I can't check institutions.

addon/adapters/osf-adapter.js
```diff
@@ -10,14 +11,82 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+        if (dirtyRelationships.length) {
+            promises = dirtyRelationships.map(relationship => {
+                var url = this._buildRelationshipURL(type.modelName, snapshot.id, snapshot, 'updateRecord', query, relationship);
+                var requestType = snapshot.record[relationship].meta().options.updateRequestType;
+                return this.ajax(url, requestType || 'PATCH', {
+                    data: this._relationshipPayload(store, snapshot, relationship)
+                }).then(() => snapshot.record.clearDirtyRelationship(relationship));
+            });
+        }
+        if (Object.keys(snapshot.record.changedAttributes()).length) {
```
samchrisinger: @abought no rush looking at this, but a second set of eyes on this promise resolution would be appreciated.

abought: This looks better- thanks for fiddling! We'll watch for edge cases, especially once the unit tests come along.

 My checklist:
- [x] Takes into account the weird behavior of `allSettled` (resolution: skips the return value for relationships)
- [x] Only returns data for the base record- isn't trying to merge in data for related records of a different type
- [x] Doesn't return data to update if relationships are the only thing updated
- [x] Resolves update issue reported here: https://github.com/CenterForOpenScience/ember-osf/pull/47#issuecomment-222249597 (an unexpected bonus in functional testing!)

abought:
When I try to affiliate an institution with a node (that is already affiliated), I get a JS console error: 

`ember.debug.js:32116 Error: allSettled must be called with an array`

samchrisinger:
@abought hopefully this ^ is addressed via https://github.com/CenterForOpenScience/ember-osf/pull/47/commits/bf6f82f79cf9873ff1eca043b1f832cf7fba6d97

abought:
> @abought hopefully this ^ is addressed via bf6f82f

Still seeing it. Here are the steps to reproduce I sent Sam via flowdock:

1. Affiliate a node on staging with cos
2. Reload the page (because hey why not)
3. Enter the same institution ID (“cos”) in the box again and click affiliate
4. Look at console and see JS errors


After this is resolved we'll check to see if Dawn's issue is resolved; otherwise, this is the last blocker on my radar for this PR for now.

addon/adapters/osf-adapter.js
```diff
@@ -85,8 +85,10 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
                 return this._super(...arguments).then(response => Ember.RSVP.allSettled(promises).then(() => response));
             }
             return this._super(...arguments);
-        } else {
+        } else if (promises) {
             return Ember.RSVP.allSettled(promises).then(() => null);
+        } else {
+            return new Ember.RSVP.Promise((resolve) => resolve(null));
```
abought: It seems kind of strange to me that we have to check for this- shouldn't `updateRecord` already know the model is clean and do nothing? (is this a sign that we're skipping some basic machinery of the method?)

PR #49, opened 2016-05-31
Bulk Request (specifically for hasMany relationships)

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-42

# Purpose
Allow bulk requests for hasMany relationships.

### DIFF
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -11,7 +11,6 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
     authorizer: 'authorizer:osf-token',
     host: config.OSF.apiUrl,
     namespace: config.OSF.apiNamespace,
-    isBulk: false,
     buildURL() {
         // Fix issue where CORS request failed on 301s: Ember does not seem to append trailing
         // slash to URLs for single documents, but DRF redirects to force a trailing slash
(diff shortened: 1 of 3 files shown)


addon/adapters/osf-adapter.js
```diff
@@ -42,8 +43,20 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
             var serializer = store.serializerFor(relationType);
             if (relationMeta.kind === 'hasMany') {
                 // A hack, since we'd have to use a bulk requests to send a list; TODO remove [0]
-                serialized = snapshot.hasMany(relationship).filter(record => record.id === null).map(record => serializer.serialize(record))[0];
-                delete serialized.data.relationships;
+                serialized = snapshot.hasMany(relationship).filter(record => record.id === null).map(record => serializer.serialize(record));
+                if (serialized.length > 1) {
+                    this.isBulk = true;
```
abought: I'm not sure about ember adapters, but controllers are singletons- a single instance of a class that handles all requests.

Because of that, adding state to an ember class is a dicey proposition; it's hard to be 100% sure it won't persist across different sessions. I'll take a look at the method internals to see if there's a better way to pass this flag around (like in the options object that's a third argument to `ajaxOptions`) without preserving state.

addon/adapters/osf-adapter.js
```diff
@@ -42,8 +43,20 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+                    serialized = {
+                        data: serialized.map(function(record) {
+                            var data = record.data;
+                            delete data.relationships;
+                            return data;
```
abought: Thinking ahead, do we intend to support other kinds of bulk requests (besides relationships)?

A typical ember pattern is to offer different methods for each behavior- kind of like how serializers offer `normalizeSingleResponse` vs `normalizeArrayResponse`. Looking at this `_relationshipPayload` method, I think we should do something similar for bulk requests, so that it is possible to override single or bulk behavior independently, without having to change both.

abought:
Quick pass complete. Three requests, as discussed with Henrique- I'll pick up again once those are ready.

PR #50, opened 2016-06-01
Feature/file manager v2 [EOSF-54]

aaxelb:
## Ticket
[EOSF-54](https://openscience.atlassian.net/browse/EOSF-54)
## Purpose
The previous `file-manager` implementation was a leaky encapsulation. It would leave the Ember store strangely dirty and wasn't fully honest when making promises.

With this change, promises returned by `file-manager` don't resolve until updated file info has propagated through the OSF and into the Ember store, so a file action can be imagined as a single, atomic operation (more or less), and the multiple moving parts are completely hidden behind this one interface.

## Notes for Reviewers
You can test this manually using the crude file view in the dummy app. If you add/delete/rename/update/move files and folders, verify that the displayed file info and file tree are updated appropriately (or inspect the Ember store, if you'd rather). Since `file-manager` no longer pushes any data into the store, those updates come from the API and indicate everything is working as intended.

Also, after each action that changes a file model, there should be an API request with `cachebypass={nonce}` in the query string (two including the OPTIONS request, cortesy of CORS). If anything else gets `cachebypass`, that's a problem.


tests/factories/file.js
```diff
@@ -3,25 +3,33 @@ import faker from 'faker';
 
 FactoryGuy.define('file', {
     default: {
-        name: () => faker.lorem.word() + '.txt',
+        name: () => faker.lorem.words(1) + '.txt',
```
abought: I know that `lorem.word` is available in newer versions of faker. Am I correct in assuming this is a workaround for old version issues?

addon/services/file-manager.js
```diff
@@ -1,43 +1,69 @@
...
+     * Upload a new version of an existing file.
+     *
+     * @method updateContents
+     * @param {file} file A `file` model with `isFolder == false`.
+     * @param {Object} contents A native `File` object or another appropriate
+     * payload for uploading.
+     * @return {Promise} Promise that resolves to the updated `file` model or
+     * rejects with an error message.
+     */
     updateContents(file, contents) {
```
abought: In the future, we may want to provide additional options to all public-facing methods, without changing the method signatures. Among the uses for these options, we might want to control internal private methods in some way, without having to resort to shared state to make additional fields accessible.

To give us room to grow, would it help to add a last argument (`options = {}` or somesuch) to all public facing methods? I'll leave it to you to decide whether the options argument would contain separate fields, eg for AJAX request configuration vs generic method configuration.

You may also want to think whether the `options` dict is passed around from the public method to the private ones- you can look at how ember does it for its own internal methods as a source of inspiration. (the `options` dict recently helped save us from some shared state with an adapter PR, for example)

addon/services/file-manager.js
```diff
@@ -141,25 +244,65 @@ export default Ember.Service.extend({
                 headers: {
                     Authorization: `Bearer ${accessToken}`
```
abought: It seems that this would bake the authorization mechanism into a service.

In the near future, we may want to make authorization type configurable, so that we (COS) can talk to the OSF via cookies. Will waterbutler always be using a token, or will the service need to support cookie auth too?

addon/services/file-manager.js
```diff
@@ -1,43 +1,69 @@
...
+        let p = this._waterbutlerRequest('PUT', url, params, contents);
+        return p.then(() => this._reloadModel(file, (model) => {
+            if (model.get('dateModified') > lastModified) {
+                return model;
+            }
+            return false;
+        }));
     },
 
     checkout(/*file, user*/) {
```
abought: Hmm. This is an extremely tough call.

I presume that users would still have the option of using the model field directly if they really wanted?

As long as this is simply part of a more cohesive interface, this seems reasonable. If we start hacking ember's internals to disable the normal file model semantics, then we can start worrying. 🐙

tests/factories/file.js
aaxelb: I was getting errors with `lorem.word`, so I guess I have an old version? I have `3.1.0` installed, but [it looks like](https://github.com/johnotander/ember-faker/blob/master/bower.json#L16) `ember-faker` depends on `~2.1.0`... could that be messing with things? I am consistently bewildered by versioning issues.

abought: Yeah, that's odd- according to `bower ls` and our bower.json, the default is to install 3.0.0-3.0.1. 

But according to `bower_components/Faker/package.js`, I have 2.1.1 locally. Usually bower should trigger a conflict in that case... I've found their dependency resolution mechanisms painful to contemplate in the past.

`words(1)` isn't a hassle; it's just that the newer fakers have some syntactic sugar for generating filenames. Ah well.

addon/services/file-manager.js
```diff
@@ -141,25 +244,65 @@ export default Ember.Service.extend({
...
+     * @private
+     * @param {Object} model Ember model to reload.
+     * @param {Function} isFresh Function that is called with the reloaded
+     * model from the OSF API. If the model is actually still stale, should
+     * return `false`. Otherwise, return the fresh model.
+     * @return {Promise} Promise that resolves to an updated `file` model or
+     * rejects with an error message.
+     */
+    _reloadModel(model, isFresh) {
+        const maxTries = 32;
```
abought: Can you clarify the purpose of the interval? 

In particular: 0.256 seconds isn't a long time to wait between attempts, and 32 retries is an extremely large number if we're talking about server traffic. 

If you're waiting for an event to occur, never use time as a proxy; tie the behavior to resolution of a promise instead. But I'm guessing the issue is getting waterbutler to register data has changed on its end- if that's the case it may be time to ping Fitz again for a better workaround.

addon/services/file-manager.js
```diff
@@ -141,25 +244,65 @@ export default Ember.Service.extend({
...
-        let attr = data.data.attributes;
-        attr.materializedPath = attr.materialized;
-        attr.dateModified = attr.modified;
+    /**
+     * Get an updated `file` model from the OSF API.
+     *
+     * @method _reloadModel
+     * @private
+     * @param {Object} model Ember model to reload.
+     * @param {Function} isFresh Function that is called with the reloaded
```
abought: Minor nitpick, but can we rename this to something that sounds more like a function, and less like a boolean?

addon/services/file-manager.js
```diff
@@ -141,25 +244,65 @@ export default Ember.Service.extend({
                 headers: {
                     Authorization: `Bearer ${accessToken}`
                 },
-                success: (data) => resolve(data),
-                error: (_, __, error) => reject(error)
+                error: (_, __, error) => reject(error),
```
abought: What is the advantage of using resolve and reject this way? Shouldn't `Ember.$.ajax` already be returning a promise, without creating another one? (or at least a promise-like "deferred" object)

If possible, use promise semantics rather than `success` and `failure` callbacks- it's a bit more consistent with Ember's interface throughout, and makes it easier to reason about chained events.

http://api.jquery.com/jQuery.ajax/#jqXHR

abought:
Pausing pass for now. The big items (in decreasing order of blockage):

- Updating the API to support user-provided options
- The usage of intervals/retries (suggests we should plan a long-term fix, and assess whether the mitigation strategy would affect the dev-facing API)
- Cleanup (including promise usages in cases where return values would be affected)
- Avoid baking the authorization method (tokens) into the helper functions if possible (eg perhaps by leveraging ember-simple-auth authorizers??) This is going on the assumption that OSF WB uses cookies; feel free to check that assumption

abought:
I'd move the ticket back if there was one. (for tracking purposes, that might not be a bad idea)

More generally, the basic API of methods seems reasonable and well thought out; I especially like that it minimizes assumptions by using the API data as authoritative source for links whenever possible. 

A few minor points of cleanup, and I may stop by to ask about WB capabilities to see if the current scheme will support extending as needed. (without changing method signatures)

abought:
I've done a very light look tonight. Requested changes appear to have been made, and the functionality I could test locally is working with the new celery priority queues. Assigning to Steve as he has requested to take a look (per MattF).

I could not test inbound mailgun hooks (like mailing list event logs). If there aren't yet official QA testing notes, I encourage you to write some up, and particularly to include anything that might trigger inconsistency between mailgun and the OSF- I'd feel much better if that was included in the testing! The need to keep data consistent with the external service is the big challenge of this PR (esp if there's a case where the celery task may not be responsible for updating status flags).

Lastly: I know we've probably committed to the design, but I'd still have preferred namespacing the mailing list names somehow (eg `project-abc12@osf.io`) to avoid the possibility of collisions with official email addresses... it would make the origin of the emails more obvious, and avoid the current very tight coupling between this feature, static OSF addresses (like `support`), and the relatively opaque GUID blacklist process.

abought:
Sorry for the crosstalk- I accidentally posted a mailing list wrap up comment on this ticket. Removed.

addon/services/file-manager.js
aaxelb: Yeah, that's a good idea, I could imagine that coming in handy for adding functionality without messing with the interface or adding state. I'll do that.

addon/services/file-manager.js
aaxelb: Good point -- this should definitely be configurable, I'll hook it up to the existing setup

addon/services/file-manager.js
aaxelb: I found that reloading the model immediately after the waterbutler request returns would usually-to-always yield the old, pre-waterbutler-action file info, but trying again after a little wait would work as expected. Instead of asking someone about this, I accepted and worked around it with the `isFresh` function to check whether it needs to try reloading again.

This is, of course, extremely hacky. I chose powers of two because it made the hackiness ever-so-slightly more palatable. If this reflects a solvable server-side problem, that would simplify file-manager nicely. I'll verify that I wasn't hallucinating and can reproduce the behavior, then annoy Fitz.

addon/services/file-manager.js
aaxelb: If it still exists after the other changes, will do.

addon/services/file-manager.js
aaxelb: It does return a deferred object, but since that has a slightly different interface from  `Ember.RSVP.Promise` (there's no `deferred.catch`), I didn't want to expose it to the outside. Wrapping it in an extra Promise seemed like a small sacrifice for consistency.

That said, I can see how it would be cleaner to use the Deferred semantics than `success` and `failure`, within the wrapping promise.

abought:
I'll pick this up for review after lunch. @aaxelb , is there a tracking ticket in JIRA for this work? If so, could you please link/assign me to it for CR?

addon/services/file-manager.js
```diff
@@ -1,133 +1,266 @@
...
-    // For actions that create entities (addSubfolder, uploadFile, copy),
-    // this is a problem, because we need the OSF info (guid, WB links)
-    // before letting the user do anything to their new file. At the moment,
-    // after you create a file, you should go get a cup of coffee, come back,
-    // and hard reload.
-
-    ///////////////////// File-only actions /////////////////////
-
-    getContents(file) {
+    isReloading: false,
```
abought: Can you explain the purpose of this flag?

In general, we'd like to try to avoid putting state on services- AFAIK these are singletons, which means that different files requests made in sequence that use the service will have the same underlying shared state.

See note below for a sample of where this could be problematic.

aaxelb:
@abought I'm just finishing up a commit that makes file-manager use ember-simple-auth instead of hard-coding the auth header. Should be up shortly, as soon as I figure out why this one test is failing...

addon/services/file-manager.js
```diff
@@ -136,30 +269,46 @@ export default Ember.Service.extend({
...
+     * @method _reloadModel
+     * @private
+     * @param {Object} model Any Ember model with a `reload` method.
+     * @return {Promise} Promise that resolves to the reloaded model or
+     * rejects with an error message.
+     */
+    _reloadModel(model) {
+        this.set('isReloading', true);
+        return model.reload().then((freshModel) => {
+            this.set('isReloading', false);
```
abought: As mentioned above, we'd like to avoid using shared state when possible. One way I can see this being a problem: if `model.reload()` fails, it seems that `then` will never be fired, and `this.isReloading` would never be set back to `false`.

Essentially, all future users of the service might get trapped in the wrong behavior because of a single failed request.

Don't `DS.Model` instances have an `isLoading` attribute on them? Would that do what you need? 
http://emberjs.com/api/data/classes/DS.Model.html#property_isLoading

addon/services/file-manager.js
aaxelb: Yeah, that's used so that when reloading file models immediately after a waterbutler action, we can [bypass the cache](https://github.com/CenterForOpenScience/ember-osf/pull/50/files#diff-060de0719e4a89c276e054713a26fe79R9) and be guaranteed the response is up to date. This was my less-hacky solution than trying multiple times.

If you have an idea for a better approach, I'd be happy to hear

addon/services/file-manager.js
aaxelb: I don't think `isLoading` is enough -- we don't want to bypass the cache on every file reload, only the ones immediately following a file create/update/delete.

But you're absolutely right that `isReloading` should be set back to `false` on failure, good catch.

abought: I'll try to think about alternatives when I review this in depth later, but this is only one of many ways that shared state on a singleton could come back to bite us. If we can get away without the instance variable on the service, that would be strongly preferred. Just setting `isReloading` back doesn't change the fact that the variable exists.

For example: what happens if I make two different file requests in rapid succession? Would the second request pick up on the `isReloading` state belonging to the first request, and trigger a different behavior than intended? That would be very hard to debug.

The key is that state is being shared among unrelated requests, and not just across time for one particular use case.

abought: And on a more positive tone: I really will try to think of alternatives when I pick this up- and would be happy to sit down and discuss logic if you want to brainstorm today. Just let me know what I can do to help.

aaxelb: That's true, any other file requests sent before the forced reload returns would also get `bypasscache=foo` added to the url, and that is far less than ideal. 

I was originally hoping that I could just pass an extra argument to `reload`, but it's frustratingly opaque, hence this screwy workaround.

addon/services/file-manager.js
samchrisinger: Could this be a hash rather than just a bool?

```
{
    <file_id>: Boolean
}
```

Might require some logic to be shifted from FileCacheBypassMixin#ajaxOptions to #buildUrl

aaxelb: I like the hash idea, but I don't think we can use `buildUrl` -- I originally put the bypasscache thing there, but it didn't work for reloading the list of files in a folder (`folder.get('files').reload()`), because it replaces the built URL with the relationship link.

I assume there's a way to get the `files` relationship link from a given `file` object, file-manager could have a whitelist of urls that should bypass the cache. Hackier than using the file id, but better than `isReloading`

addon/services/file-manager.js
```diff
@@ -1,165 +1,318 @@
...
+        options.data = contents;
+        if (!options.query) {
+            options.query = {};
+        }
+        options.query.name = name;
+        options.query.kind = 'file';
 
-    ///////////////////// File and folder actions /////////////////////
+        let p = this._waterbutlerRequest('PUT', url, options);
+        return p.then(() => this._getNewFileInfo(folder, name));
```
abought: Isn't this data returned from the server as the result of a PUT request? 

Am I correct in assuming that we are sending two requests as an alternative to manually updating the store?

addon/mixins/file-cache-bypass.js
```diff
@@ -0,0 +1,16 @@
...
+export default Ember.Mixin.create({
+    fileManager: Ember.inject.service(),
+
+    ajaxOptions() {
+        // When reloading file models after a file action, bypass the cache
+        let options = this._super(...arguments);
+        if (this.get('fileManager.isReloading')) {
+            let prefix = options.url.indexOf('?') === -1 ? '?' : '&';
+            let nonce = Date.now();
+            options.url += `${prefix}cachebypass=${nonce}`;
```
abought: A docstring here might be helpful for posterity.

As clarified by Abram: this is not a documented behavior triggered by specific API code. Rather, adding *anything* unique to the URL will force varnish to bypass the cache.

addon/services/file-manager.js
abought: As discussed in person, the issue here is that Ember enforces separation of concerns, eg `model.reload()` is assumed to have no knowledge on URLs.

Two options: 
- Track a temporary state parameter on the model / snapshot (accessible to adapter)
- Key the record to a hash / object (rather than a single boolean that stores shared state across all file requests)

addon/services/file-manager.js
```diff
@@ -1,165 +1,318 @@
...
+    _waterbutlerRequest(method, url, options = {}) {
+        if (options.query) {
+            let queryString = Ember.$.param(options.query);
             url = `${url}?${queryString}`;
         }
-        let sessionData = this.get('session').get('data').authenticated;
-        let accessToken = sessionData.attributes.accessToken;
+
+        let headers = {};
+        this.get('session').authorize('authorizer:osf-token', (headerName, content) => {
```
abought: This about tracks with what I saw of ember simple auth- not as supportive of cookies as we would like, but it may be what we're stuck with. Thanks.

abought:
I don't know that I'll ever love the waterbutler API, but from the looks of it, I might not mind using this wrapper- and that's no small thing.

Returning to @aaxelb for relatively light changes around the `isReloading` parameter. No other major blockers noted.

aaxelb:
I removed `isReloading` and replaced it with a (private) hash of URLs and a public method to check whether a URL is in the hash. It gets the URL for reloading individual files from `file.get('links.info')` and the URL for reloading `folder.files` relationships from `folder.get('files').get('content.relationship.link')`, which uses some private properties on the `hasMany` relationship. It's hacky, but should work reliably.

I couldn't use just a hash of file/folder IDs, like we talked about, because it wouldn't cover reloading the contents of a provider's root folder.

I also added some docstrings explaining this whole hack.

addon/services/file-manager.js
aaxelb: We talked about this in person, but for anyone else: The waterbutler entity returned for newly created files doesn't include the file ID used by the OSF API. In order to use the new file in any meaningful way, we need to get its metadata from the API by listing the contents of its parent folder.

For other (non-create) operations, we might technically have enough information to manually update our models to what they should be, but I'd still prefer reloading the model through the API, to be sure our models reflect what actually is.

addon/services/file-manager.js
```diff
@@ -245,6 +243,64 @@ export default Ember.Service.extend({
...
+     * @method _reloadModel
+     * @private
+     * @param {Object} model `file` model or a `files` relationship
+     * @return {Promise} Promise that resolves to the reloaded model or
+     * rejects with an error message.
+     */
+    _reloadModel(model) {
+        // If it's a file model, it has its own URL in `links.info`.
+        let reloadUrl = model.get('links.info');
+        if (!reloadUrl) {
```
abought: Purely for future reference, this can be written more concisely using lazy evaluation, as:

`let reloadUrl = model.get('links.info') || model.get('content.relationship.link');`

This sort of pattern is particularly common in JS, where (transpiling aside) people are trying to write more concise code to save bandwidth. Fortunately, in this case, it's not a big difference in readability. :)

abought:
Although the testing notes recommend uploading files, clicking the "Add folder" and "Upload file" buttons does not appear to do anything; is that intentional?

Sample route:
http://localhost:4200/nodes/m9pv6/files/osfstorage

abought:
Trying to upload a new file on top of an existing file is also giving 500 server errors, so not sure what's happening tonight. Feel free to nudge me in the morning to try review again, so we can get this moving. :)

abought:
Working this AM for me, so moving forward. Ready to merge, barring any feedback/questions from @samchrisinger .

PR #31, opened 2016-05-17
[WIP] Allow login with OSF cookie

abought:
Do not merge; in progress. Created for tracking purposes.

## Ticket
https://openscience.atlassian.net/browse/EOSF-37

# Purpose
Allow ember-osf to connect with a cookie. This would be used solely for COS internal use, as we will generally discourage third-party applications from using the OSF cookie.

# Notes for Reviewers
To be written

# TODO
- Test against OSF backend
- Finish authorizer/ authenticator
- Get/write tests
- Figure out why ember simple auth is ignoring configurable login route

### Routes Added/Updated
- Add a sample cookie login route to the dummy app purely for testing purposes.
- Make authorizor/authenticator configurable

PR #54, opened 2016-06-02
Cookie auth

abought:
## Ticket
Authentication: https://openscience.atlassian.net/browse/EOSF-37
Mock endpoint to set and use cookies: https://openscience.atlassian.net/browse/EOSF-48

Replaces and extends #31 .

# Purpose
Allow ember-OSF to support cookie authentication, when running inside the OSF app. This would be used solely for COS internal use, as we will generally discourage third-party applications from using the OSF cookie.

# Notes for Reviewers

## Deployment considerations
**This renames the login controllers (to be specific about tokens vs cookies). This will break downstream consuming apps unless they update!**

## Testing this feature
First, make sure you are logged out of localhost OSF (this uses the same cookie!)

This is somewhat unpleasant to get set up. It involves the following changes:
1. In ember-cli-build, un-comment these lines. They are responsible for adding an asset path prefix to the built assets, so that ember can be served from within the OSF. Build process improvements are a separate ticket and normally we wouldn't be building an addon as an app.
```
        fingerprint: {
            prepend: 'static/ember/'
        },
```
2. In config/local.yml, change your redirect URI to go to the special cookie page (not the token login route): `REDIRECT_URI: http://localhost:5000/cookielogin/`
3. From your ember-osf root directory run the following command to build the ember files, use cookie auth, and put assets inside an OSF static folder:
`AUTHORIZER=cookie BACKEND=local ember build --output-path ../osf.io/website/static/ember/ --environment=production --watch`
4. To serve the ember app, the OSF must have a route matching one in your ember app. You can use [this branch](https://github.com/abought/osf.io/tree/demo/ember-from-osf), which lets you use `/cookielogin` and `/nodes` ember demo app routes from `localhost:5000`
https://github.com/abought/osf.io/tree/demo/ember-from-osf
5. Go to `http://localhost:5000/cookielogin`. Log in on the OSF via the button in the main page area. It will return you to the page and auto-redirect to the nodes page.
  - Do not use the navbar button popup. It causes some weird security warnings that need to be investigated.

## Known issues and vexations
Doing things by AJAX requests on localhost imposes a legion of new challenges.
1. You have to use the page login button. The navbar popup fails due to domain-related security issues
2. The logout process redirects you to a new page. This isn't allowed with CORS, so we need to change window location in order to completely destroy the session. (the navbar doesn't magically reload)
3. The prepend option is intended for development only. This means that rebuilds are quite slow. (10-20 seconds) This is really because we're trying to treat the dummy app as a real thing, and it makes the demo needlessly complicated to use.
4. The navbar is not very configurable. 
5. Every single AJAX call needs to manually add configuration
  - We can probably write a helper function to set up the ajax call, and make it work nicer with configurable authorizers.

## OSF cookie auth flow
1. A user logs in via CAS
2. CAS redirects the browser to a specified URL back on the OSF, appending a query parameter `?ticket=value`. Because we serve from within flask, the ember app doesn't need to handle the exchange mechanism- the cookie gets set by flask as it serves up the HTML page
  - This leads to a weird situation where we need to force ember-simple-auth to try logging in, whenever the cookielogin loads. It's pretty ugly. Really, we are not using ember-simple-auth very much for this.
3. The OSF responds to this ticket parameter by trading it to CAS for credentials on the backend (handled in `before_request`), and sets an OSF cookie as part of the response
4. The OSF cookie than gets sent along with every request. In theory we can make this happen automatically, but I had trouble talking to the API if I failed to set `withCredentials` on the xhr object.

## How to verify
- You should be able to see pages that rely on user-specific data, such as `/nodes`, after login.
- The request headers to the API endpoint should not include bearer tokens.
- The network tab (eg in chrome) for dev tools should show that the request (and headers) includes cookies
- Ensure that manual ajax queries also work with cookies- primarily this will affect things that use the file manager
- Log out and confirm it takes you to another page (flask `/goodbye` route for now)

## Routes Added/Updated
- Adds a route to demonstrate cookie usage: `/cookielogin`, with corresponding mixins


addon/authenticators/osf-cookie.js
```diff
@@ -0,0 +1,54 @@
...
+        return this._test().fail(this.invalidate);
+    },
+
+    /**
+     * Send a request to the flask application to trigger invalidation of session remotely
+     * @method invalidate
+     */
+    invalidate() {
+        // Can't do this via AJAX request because it redirects to CAS, and AJAX + redirect = CORS issue
+        window.location = `${config.OSF.url}logout/`;
```
samchrisinger: The CAS exposes a logout endpoint too: `(.*-)?accounts.osf.io/logout`... would that make more sense here?

addon/utils/ajax-helpers.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Performs any additional authorization config required, for the configured authorization type.
+ * Allows manual AJAX requests to be authorization-agnostic.
+ *
+ * Primarily used to set XHR flags on manual AJAX requests, for cookie based authorization.
+ * @method ajaxAuth
+ * @param {Object} options
+ * @returns {Promise}
+ */
+function ajaxAuth(options) {
```
samchrisinger: This is very minor, but the name `ajaxAuth` doesn't really convey this method's purpose super well to me. Something like `autheticatedAjax` might be clearer? Thoughts here?

ember-cli-build.js
```diff
@@ -11,6 +11,18 @@ module.exports = function(defaults) {
                 'addon/styles'
             ]
         },
+
+        // FIXME
```
samchrisinger: @abought can we target this block for removal as a part of your current work?

samchrisinger:
@abought looks great, just a couple questions then rtm

addon/authenticators/osf-cookie.js
abought: In principle, the OSF keeps track of its own sessions. Hitting CAS alone would not guarantee that the session cookie was removed/invalidated. Here's the method the OSF uses to logout- it hits CAS, but also does other things.
https://github.com/abought/osf.io/blob/a9054b02310f31223933a9f62116574176fce1d5/framework/auth/views.py#L277-L277

...Come to think of it, I'm not sure whether/how that causes the cookie to be actually invalidated server side (if someone presents it as a bearer instrument later). But it does at least remove the cookie from the user's computer.

samchrisinger: 👍 makes sense

ember-cli-build.js
abought: By current work, do you mean picking up EOSF-88 (OSF consuming app)?

samchrisinger: yes

addon/utils/ajax-helpers.js
abought: I'd been going for similarity to jquery method names, but this is an easy change to make. 

(if someone quibbles about authentication vs authorization, we can always rename the method to `zxcvbnm` later)

abought:
Changes pushed.

Incidentally, for checking docs, there's a handy development mode in YUIDoc:
`yuidoc --server 3000`

Lets you refresh the browser and quickly see changes made to pages at http://localhost:3000

Warnings (syntax/ build errors) will get printed to the console, if you don't want to wait to look at `docs/data.json : warnings[]` after a full build.

addon/authenticators/osf-cookie.js
abought: Just to wrap up a dangling thought...

Oh, good, the OSF does also remove its own session during the flask application logout process. :)
https://github.com/abought/osf.io/blob/a9054b02310f31223933a9f62116574176fce1d5/framework/auth/__init__.py#L54-L54

PR #59, opened 2016-06-08
Bulk requests for node hasMany relationships, Contributors + Components 

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-62

# Purpose
Allow for bulk updates, especially for node hasMany relationships being updated when `node.save()` is called.

## Routes Added/Updated
Node detail has bulk add contributors for testing, and contributor settings update is now a bulk request that hits /node/contributors by calling save over hitting the individual contributor endpoints.


addon/adapters/osf-adapter.js
```diff
@@ -68,21 +37,64 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
...
+            serialized = serialized[0];
+        }
+        return serialized;
+    },
+    _handleManyRequest(store, type, snapshot, query, relationship, serializer) {
+        var relationMeta = snapshot.record[relationship].meta();
+        var serialized = snapshot.hasMany(relationship).filter(each => each.id === null || Object.keys(each.changedAttributes()).length).map(each => serializer.serialize(each));
+        var options = Ember.$.extend({},
+            {
+                requestType: () => 'PATCH',
```
samchrisinger: Is there a reason this is a function?

hmoco: to allow for relationships that use more than one requestType, this needs to be called.

addon/models/osf-model.js
```diff
@@ -33,24 +33,35 @@ export default DS.Model.extend({
...
-                    var key = `_dirtyRelationships.${rel}`;
-                    this.set(key, !Ember.isEmpty(this.get(key)));
-                    // Checks is_embedded flag set in osf-serializer.
-                    if (this.get('relationshipLinks') && this.get('relationshipLinks')[rel] && this.get('relationshipLinks')[rel].is_embedded) {
-                        this.set(key, false);
-                        this.get('relationshipLinks')[rel].is_embedded = false;
+            } else if (meta.kind === 'belongsTo') {
+                let relation = this.belongsTo(rel).belongsToRelationship;
+                if (relation.record.isLoaded()) {
+                    var record = relation.members.list[0];
```
samchrisinger: Do we need to use members or canonicalState?

hmoco: members.list here is what we want, since this will always exist provided the relationship has data.

samchrisinger: The filter above uses canonicalState. Is that one correct?

hmoco: Yep (for that one it actually doesnt matter, but canonicalState makes sense). So there, you are making two different checks:
- Have the current members (canonical) been modified?
- Have we removed or added members (comparison between members and canonical).
For the belongsTo, we want to make sure that it grabs something, so members making sense (as if that member had just been added, it would not be in canonical)

samchrisinger: :+1:, thanks

hmoco: The one edge case is if someone adds an existing object, and removes another previous to a save, without changing any of their attributes, which we can avoid by iterating through both canonical and members and doing a set comparison of the ids, but I don't know if that is necessary

addon/serializers/contributor.js
```diff
@@ -8,7 +8,7 @@ export default OsfSerializer.extend({
         serialized.data.relationships = {
             users: {
                 data: {
-                    id: snapshot.record.id,
+                    id: snapshot.record.get('id') || snapshot.record.id.split('-')[1],
```
samchrisinger: Can you explain this? I read:
> get the id. If id is null, try to get the id perform a split

hmoco: explain why I changed it? This is your change, I made it .get('userId')

PR #60, opened 2016-06-09
Feature/Tags Widget [EOSF-59]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-59

# Purpose

Adds tags component and demonstrates use for both node tags and file tags in the dummy app.

# Notes for Reviewers

addATag and removeATag methods on the controllers are a little weird.  I was having a hard time getting `snapshot.record.changedAttributes` to update.  For adding a tag, the only way I could get it to change was if I made a copy of the current tags (that did not reference the original), pushed the new tag onto the list, and then set the tags to this list.  If I pushed the new tag onto the current tags and then tried to set, changedAttributes would be an empty object.

Also, adding/removing tags from files are currently creating an extra log (checked_in), which is an OSF problem, to be fixed here https://github.com/CenterForOpenScience/osf.io/pull/5820.  The bigger problem is that for our API PATCH requests we are serializing everything.  Our API requests should just send the pieces we want to change ticket here https://openscience.atlassian.net/browse/EOSF-61

### DIFF
--- a/addon/components/tags-widget/component.js
+++ b/addon/components/tags-widget/component.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
+    layout,
+    tags: [],
+    actions: {
+        addATag(tag) {
+            this.attrs.addATag(tag);
+        },
+        removeATag(tag) {
+            this.attrs.removeATag(tag);
+        }
+    }
+});
--- a/addon/components/tags-widget/component.js
+++ b/addon/components/tags-widget/component.js
@@ -6,10 +6,10 @@ export default Ember.Component.extend({
     tags: [],
     actions: {
         addATag(tag) {
-            this.attrs.addATag(tag);
+            this.sendAction('addATag', tag);
         },
         removeATag(tag) {
-            this.attrs.removeATag(tag);
+            this.sendAction('removeATag', tag);
         }
     }
 });
--- a/tests/dummy/app/controllers/nodes/detail/files/provider/file.js
+++ b/tests/dummy/app/controllers/nodes/detail/files/provider/file.js
@@ -5,6 +5,21 @@ export default Ember.Controller.extend({
         reloadFiles() {
             this.transitionToRoute('nodes.detail.files.provider',
                 this.get('node'), this.model.get('provider'));
+        },
+        addATag(tag) {
+            var file = this.get('model');
+            var currentTags = file.get('tags').slice(0);
+            Ember.A(currentTags);
+            currentTags.pushObject(tag);
+            file.set('tags', currentTags);
+            file.save();
+        },
+        removeATag(tag) {
+            var node = this.get('model');
+            var currentTags = node.get('tags').slice(0);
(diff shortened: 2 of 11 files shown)


addon/components/tags-widget/component.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
+    layout,
+    tags: [],
+    actions: {
+        addATag(tag) {
+            this.attrs.addATag(tag);
```
samchrisinger: Nice job on the DDAU approach. I think we'll want to use `sendAction` here (http://emberjs.com/api/classes/Ember.Component.html#method_sendAction), e.g.:

`this.sendAction('addATag', tag)`

tests/dummy/app/controllers/nodes/detail/files/provider/file.js
```diff
@@ -5,6 +5,21 @@ export default Ember.Controller.extend({
         reloadFiles() {
             this.transitionToRoute('nodes.detail.files.provider',
                 this.get('node'), this.model.get('provider'));
+        },
+        addATag(tag) {
+            var file = this.get('model');
+            var currentTags = file.get('tags').slice(0);
```
samchrisinger: I'm wondering if we could just 

`file.get('tags').pushObject(tag)` ?

tests/dummy/app/controllers/nodes/detail/index.js
```diff
@@ -19,6 +19,21 @@ export default Ember.Controller.extend({
             var bibliographic = target.checked;
             var contributorId = target.value;
             this.editedBibliographic[contributorId] = bibliographic;
+        },
+        addATag(tag) {
```
samchrisinger: Since we're doing this in two places, would it make sense to create a mixin to generalize this behavior? This could be a `TaggableMixin` that can be mixed into controllers and routes.

samchrisinger:
@pattisdr pass done, looks great so far!

tests/dummy/app/controllers/nodes/detail/files/provider/file.js
pattisdr: This was the first thing I tried.  I remember it saying that pushObject was not a function, so I changed it to `Ember.A(file.get('tags')).pushObject(tag)`.  But this results in `changedAttributes` being an empty object.  So no PATCH request is sent.

samchrisinger: Ok. @hmoco's latest work may fix this. I'll add a TODO to check into this when his changes are ready.

pattisdr:
@samchrisinger, requested changes made!

samchrisinger:
@pattisdr looks good. I would like to see some [integration tests added](https://openscience.atlassian.net/browse/EOSF-63), but won't let that be a blocker for merge.

pattisdr:
I did add one test making sure a tag was rendered properly.  What else would you like tested? The actions?

samchrisinger:
Oh, I didn't see that one. But yes, it would be good to _unit_ test the actions.

PR #61, opened 2016-06-10
Commenting widgets

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-20

Replaces and extends PR #41 

# Purpose
- Turn comment code on file template into a component (`comments-pane`) made out of smaller components (`comment-detail` and `comment-form`)

screen shot 2016-07-01 at 2 20 18 pm

# Notes for Reviewers
## Testing notes
- Widget should support a variety of common comment actions:
  - [x]  create
  - [x] edit
  - [x] delete
  - [x] restore
  - ~~report/unreport?~~ (separate ticket)
- It should add newest comments to the top of the list- the same sort order used by the API.
- If you are the person who made a comment, you should see edit and delete buttons
  - If you are the owner of a deleted comment, you should see a restore button (but no text)
- If you are not the owner of a comment, you should see a report button
  - If you are not the owner of a deleted comment, you should not see comment text
- Comments on different things should be kept separate, eg project comments pane should not display the comments on a file owned by that project. (even though they use the same API endpoint)

## Known issues
- Initial pageload currently limited to 10 items (pagination).
- Cannot report comments (separate PR).=
- No support for threaded conversations. Emphasis is on model functionality.
- Doesn't check whether user can comment on project.

## Routes Added/Updated
- The project detail page has a new Comments section: http://localhost:4200/nodes/f263w
- The file detail page has a new Comments section: http://localhost:4200/nodes/3t9r8/files/osfstorage/576989df0dc31001ecbd4ef3

PR #66, opened 2016-06-13
[WIP] refactoring Node actions into mixin

samchrisinger:
# Purpose

Moving actions and action listeners to callbacks promotes reusability and shareability of code. This allows for a single canonical understanding of how to interact with the ember-osf models. 

This PR attempts to consolidate some of the Node actions into a single Mixin.


tests/dummy/app/app.js
```diff
@@ -6,6 +6,7 @@ import config from './config/environment';
 let App;
 
 Ember.MODEL_FACTORY_INJECTIONS = true;
+Ember.run.backburner.DEBUG = true;
```
abought: Do we intend to keep this flag on? I'm curious how your experience fits with the warning in the docs:

From https://guides.emberjs.com/v2.4.0/configuring-ember/debugging/#toc_errors-within-code-ember-run-later-code-a-href-https-github-com-ebryn-backburner-js-backburner-js-a
> Backburner has support for stitching the stacktraces together so that you can track down where an erroring Ember.run.later is being initiated from. Unfortunately, this is quite slow and is not appropriate for production or even normal development.

samchrisinger: AFIK this will only apply to the dummy app, so I don't think it's unreasonable to have this turned on. But yeah, definitely need to make sure this doesn't affect consuming applications.

tests/dummy/app/templates/nodes/detail/index.hbs
```diff
@@ -82,48 +80,12 @@
...
-        {{radio-button name="permission" value="read" checked=contributorPermissions}} Read {{radio-button name="permission" value="write" checked=contributorPermissions}} Read + Write {{radio-button name="permission" value="admin" checked=contributorPermissions}} Administrator
-    </p>
-
-    <label> Bibliographic </label>
-    <p>
-        {{radio-button name="bibliographic" value=1 checked=bibliographic}} Bibliographic {{radio-button name="bibliographic" value=0 checked=bibliographic}} Non-Bibliographic
-    </p>
-    <button {{action "addContributor" contributorID contributorPermissions bibliographic}} class="btn btn-primary">Add</button>
+    {{contrib-manager
+    contributors=model.contributors
```
abought: Not a problem for now, but long-term, we'll have to think about how to handle paginated relationships in these components. (`model.contributors` would be only 10 results, which could be problematic for, eg, RPP)

tests/dummy/app/templates/nodes/index.hbs
```diff
@@ -4,7 +4,7 @@
     {{input type="text" value=title placeholder="Node's title" required=true}}
     <label>Description (optional):</label>
     {{input type="text" value=description placeholder="Node's description"}}
-    <p><button {{action "createNode" title description}}>Create Node</button></p>
+    <p><button onclick={{action 'createNode' title description}}>Create Node</button></p>
```
abought: Shouldn't this work without specifying `onclick`?

I've been using that syntax on some of my recent PRs, eg
https://github.com/CenterForOpenScience/ember-osf/pull/61/files#diff-0c1696451128e99e21bc7e35c6983fadR6

tests/dummy/app/components/contrib-manager/template.hbs
```diff
@@ -0,0 +1,54 @@
...
+            </td>
+            <td>
+                <select onchange={{action 'permissionChange' contrib value='target.value' }}>
+                    <option value=READ selected={{if (eq contrib.permission READ) true false}}> Read </option>
+                    <option value=WRITE selected={{if (eq contrib.permission WRITE) true false}}> Read + Write </option>
+                    <option value=ADMIN selected={{if (eq contrib.permission ADMIN) true false}}> Administrator </option>
+                </select>
+            </td>
+            <td>
+                <input onchange={{action 'bibliographicChange' contrib value='target.value' }} checked=contrib.bibliographic}} type="checkbox" name="bibliographic">
```
abought: Note for separate audit after this PR: form controls should have associated label elements

tests/dummy/app/templates/nodes/index.hbs
samchrisinger: I think so, but have a slight preference towards the explicitness on `onclick=`

hmoco:
Function review brought out a few bugs, would prefer if these pieces of functionality worked before code reviewing:
 - visiting a node's page triggers:
```
TypeError: Cannot read property 'links' of undefined
 at Class._extractEmbeds (osf-serializer.js:39)
    at Class._mergeFields (osf-serializer.js:56)
```
 - Attempting to add a component triggers:
```
Uncaught Error: Property set failed: object in path "_dirtyRelationships.contributors" could not be found or was destroyed.
```
 - Attempting to add multiple components triggers:
```
Nothing handled the action 'addChildren'. If you did handle the action, this error can be caused by returning true from an action handler in a controller, causing the action to bubble.
```

hmoco:
@samchrisinger  ^

hmoco:
Also adding institutions returns a 200 but doesn't persist.

PR #69, opened 2016-06-14
Feature/Draft Registrations Model [EOSF-11]

pattisdr:
# Ticket

https://openscience.atlassian.net/browse/EOSF-11

# Purpose

Adds draft registration model, metaschema model, and expands registration model.  Adds ability to view a node's draft registrations, and node's registrations, create a draft, update a draft, and create a registration.

TODO
- [x] Add draft registration model
- [x] Add metaschema model
- [x] Add metaschema/draft registration serializers and adapters
- [x] Display list of a node's draft registrations
- [x] Display list of a node's registrations
- [x] Add draft_read and draft_write to full_read and full_write scopes Solved: https://github.com/CenterForOpenScience/osf.io/pull/5838
- [x] Return draft_registration link under node relationships links Solved: https://github.com/CenterForOpenScience/osf.io/pull/5838
- [x] View a single draft
- [x] Create a draft
- [x] Edit a draft
- [x] Create a registration from the draft
- [x] Delete a draft

# Notes for Reviewers

- dirtyRelationships incorrectly being set when draftRegistration is updated. `initiator` and `registrationSchema`) fields are being marked as dirty.
- This might be preventing creating a registration when final update to draft required - currently can only create registration if you have not simultaneously tried to update draft on same page.
- Excessive logic in draft registration detail template - should be moved out eventually.
- Not focusing on draft registration UI for this - will need more work.
- File upload has not been implemented.
- Registration detail template should be expanded at some point.
- Veer prereg schema not being handled yet.

## Routes Added/Updated

- /nodes/detail/draft_registrations/
- /nodes/detail/registrations/
- /nodes/detail/draft_registrations/detail
- /metaschemas/


addon/models/draft-registration.js
```diff
@@ -0,0 +1,19 @@
+import DS from 'ember-data';
+import OsfModel from './osf-model';
+
+export default OsfModel.extend({
```
abought: Huzzah, new models!

We've recently started adding documentation comments to our model files, so that implementers can easily find information on what each field means (and how to use the model). Could you do the same for these new models?

These links typically point at production swagger docs, with a direct link to the section for that endpoint. (an intern is making those links work better on the swagger end). See examples:
https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/models/file-version.js#L6

tests/dummy/app/helpers/check-data-type.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export function checkDataType(resource) {
```
abought: This seems to be a validation function?

For internal use in the dummy app that's probably fine. 

As we build components, we will want to consider using an established prebuilt form validation library for this purpose rather than writing our own.

pattisdr: You're right, that will be better.  My goal for this PR was to try to quickly test the model and make sure we could create, update, delete drafts, and create registrations (testing the new API v2 functionality).   In the future, the forms for these drafts will need to be improved.

tests/dummy/app/templates/nodes/detail/draft-registrations/detail.hbs
```diff
@@ -0,0 +1,71 @@
...
+            <p> {{question.description}} </p>
+        {{/if}}
+        {{#if (eq question.type "choose")}}
+            {{#unless question.description}}
+                <p> <em> Please choose </em> </p>
+            {{/unless}}
+            <form>
+                {{#each question.options as |option|}}
+                    {{#if (eq (checkDataType option) "string")}}
+                        <p> <input onchange={{action "buildForm" value="target"}} type="radio" name={{question.qid}} value={{option}} checked={{selectedHelper option (draftTextInput model.draft.registrationMetadata question)}}> {{option}} </p>
```
abought: For future reference... although ember [doesn't](https://guides.emberjs.com/v2.4.0/templates/writing-helpers/#toc_helper-names) *require* helper names to have a dash, this syntax is generally preferred over `camelCase` helper names.

In addition to consistency, I believe the reasoning is that it's less likely to conflict with existing JS variable names of HTML tags. (at least, that's why [web components](http://webcomponents.org/articles/how-should-i-name-my-element/) insist on hyphenated names...)

abought: I won't be looking to block this PR based on UI code, but I'd be happy to pair sometime to look over alternate ways of dealing with UI. 

The use of nested helpers and validators makes it a bit hard to follow the intended logic here, but there are some alternative approaches.

tests/dummy/app/helpers/eq.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export function eq(params) {
```
abought: If we decide that we really do need truth helpers in handlebars templates, this library may be of interest:
https://www.npmjs.com/package/ember-truth-helpers

tests/dummy/app/templates/nodes/detail/draft-registrations/detail.hbs
pattisdr: Yes, I mentioned this above in the PR description.  There is way too much logic in this template!

tests/dummy/app/routes/nodes/detail/children.js
```diff
@@ -2,10 +2,7 @@ import Ember from 'ember';
...
 export default Ember.Route.extend({
     model() {
-        return this.store.query('node', {
-            filter: {
-                parent: this.modelFor('nodes.detail').get('id')
-            }
-        });
-    }
+        let node = this.modelFor('nodes.detail');
+        return node.get('children');
```
abought: Can you explain the reason for this change?

By default ember doesn't support paginating over relationships, which would limit this query as written to the first 10 results. If you need more, I can help point you to a few solutions available in our codebase.

tests/dummy/app/routes/nodes/detail/draft-registrations.js
```diff
@@ -0,0 +1,39 @@
+import Ember from 'ember';
+import permissions from 'ember-osf/const/permissions';
+
+export default Ember.Route.extend({
+    model() {
+        let node = this.modelFor('nodes.detail');
+        let drafts = node.get('draftRegistrations');
+        return Ember.RSVP.hash({
```
abought: Not a blocking comment for this PR, but...

Conventionally, Ember favors using nested resources, and it gets super-snarky if the model hook depends on two items- so using RSVP.hash in a model hook is gently disfavored.

There are a few alternative ways to handle this (like `setupController` or `afterModel`) depending on your use case.

tests/dummy/app/routes/nodes/detail/children.js
pattisdr: I wrote the original method, querying nodes that had a specific parent, which resulted in an APIv2 call to `v2/nodes/?filter[parent]={node_id}`.   The change here made an API v2 request to `v2/nodes/{node_id}/children/`.  It seems like few of the relationships even have this inverse available for querying, so I was just trying to be consistent with other patterns.  Since we have the nodes/node_id/children endpoint, it didn't seem necessary to filter all nodes on parent.

But I did not know the bit about paginating over relationships.

tests/dummy/app/routes/nodes/detail/draft-registrations.js
```diff
@@ -0,0 +1,39 @@
...
+    actions: {
+        createDraft(schemaId) {
+            var node = this.modelFor(this.routeName).node;
+            if (node.get('currentUserPermissions').indexOf(permissions.ADMIN) !== -1) {
+                var draft = this.store.createRecord('draft-registration', {
+                    registrationSupplement: schemaId
+                });
+                node.get('draftRegistrations').pushObject(draft);
+                node.save();
+                node.one('didUpdate', this, function() {
```
abought: `save` should return a promise; it may be simpler to use `.then` instead of `didUpdate` here.

abought:
Very helpful reviewer notes! 

For a PR focused on the modes, it sounds like the key thing to evaluate is whether data goes to/from the server correctly. What is the current status of the note below- and what can I do to help resolve the outstanding issue?

> dirtyRelationships incorrectly being set when draftRegistration is updated. initiator and registrationSchema) fields are being marked as dirty.

tests/dummy/app/templates/nodes/detail/draft-registrations/detail.hbs
pattisdr: Regarding the helpers, they do have dash names, but when you do ember generate helper selected-helper, when the helper is exported, it automatically makes them camelCase.  So the selectedHelper referred here, its file is app/helpers/selected-helper.js.

pattisdr:
@abought, Sam said he hoped https://github.com/CenterForOpenScience/ember-osf/pull/66 would resolve. You can still update a draftRegistration, but for the time being, extra PATCH requests are being sent to update metaschemas and the user, which is incorrect.  You can also create a registration, but you need to make sure you've already edited your metadata and saved it.

I have verified that all of the requests work - creating, updating, viewing, and deleting drafts, and creating registrations in this PR, however.

tests/dummy/app/templates/nodes/detail/draft-registrations/detail.hbs
abought: Right, but you should still be able to use the dash-cased form. That's what we do with the `elem-id` helper, and it didn't seem to require any special config to make it work.

(if ember really is supporting two names at once, that's admittedly very confusing!)

pattisdr: I see.

tests/dummy/app/controllers/nodes/detail/draft-registrations/detail.js
```diff
@@ -0,0 +1,40 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+    editedMetadata: {},
```
abought: For future reference, ember controllers are singletons, which means that a single instance might be shared between visits- that makes shared storage like this a potential source of subtle bugs.

When we do make UI improvements, we'll want to be careful to ensure that a user navigating within the app won't see state left over from a previous visit to the page. Not a blocker for this PR though.

pattisdr: Good point.

tests/dummy/app/helpers/eq.js
samchrisinger: Please go ahead and install `ember-truth-helpers` (which provides this helper and many other useful ones too)

tests/dummy/app/routes/nodes/detail/draft-registrations.js
samchrisinger: +1

tests/dummy/app/routes/nodes/detail/draft-registrations.js
samchrisinger: I agree with Andy here. We've experienced a lot of unexpected behavior when using hashes in the model hook. In this case, using `setupController` to set the `node` property on the controller is the preferred solution here.

addon/models/draft-registration.js
```diff
@@ -1,12 +1,22 @@
 import DS from 'ember-data';
 import OsfModel from './osf-model';
 
+/**
+ * Model for OSF APIv2 draft registrations.
+ * This model represents draft registration data - the initiator of the draft,
```
abought: Our API has some slightly unusual behavior, in that certain models/collections can only be accessed via relationships (`this.store.find('modelname')` would fail).

To help avoid confusion, can you please indicate in the text whether this is intended for direct querying, relationships, or both?

For maintainability reasons, we're trying to keep the list of fields/descriptions all in one place (the swagger docs). If we do want to explain fields in more detail, we should do so by calling out each field (and type) explicitly in [YUIDoc](http://yui.github.io/yuidoc/syntax/) syntax.

abought:
Review pass complete. 

Confirmed that basic viewing and saving of models works. (subject to the dirty relationship tracking stuff).

Once #66 goes in, I'd strongly encourage a followup PR to add regression tests for model behaviors, in cases where we saw bugs. Whether those would be tests of the adapter, or these specific models, is up to you.

A variety of comments were made for UI, but those won't block this PR going forward. We'll want to component-ize it for any production code.

tests/dummy/app/routes/nodes/detail/draft-registrations.js
pattisdr: Is save returning a promise or have we overridden it? I am getting `Cannot read property 'then' of undefined`.

samchrisinger: @pattisdr this should be fixed in develop. The current approach looks fine, we can iterate as we move some of this code into more reusable forms.

samchrisinger: https://github.com/CenterForOpenScience/ember-osf/commit/35a0cfd67a2f0b288acec97a2143889a6bd9bed3

abought: If it has been overridden, this should probably be fixed when we do the adapter cleanup.

(at one point we were altering what/whether `save` returned for relationship updates... in which case the solution is to carefully update our code to be consistent with Ember conventions)

pattisdr: Thanks Sam - it is fixed in develop.

PR #70, opened 2016-06-16
[R&D] "Fetch all" functionality for paginated routes

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-67

# Purpose
Demonstrate the ability to fetch all records from a paginated route.

Basic implementation for records is possible using unmodified ember-infinity, but a fetch-all function for relationships will require further customization, as well as a separate addon.

# Notes for Reviewers
## Key changes
- Add [ember-data-has-many-query](https://github.com/mdehoog/ember-data-has-many-query) to all OSF models. This is the piece that enables querying / pagination for relationships. That capability will now be available throughout the application.
  - Should we consider moving it from `devDependencies` to `dependencies` in `package.json`?

- Add `FetchAllRouteMixin` (built on lightly tweaked ember-infinity)

## Limitations
- This is written as a route-mixin. This is consistent with ember-infinity behavior, and with ember patterns that encourage data to be owned by the route.
- Ember infinity inherently supports only one infinite model per route. Therefore, this method will only inherently support one infinite fetch (model or relationship) per route.
- Due to how relationship query functions work, this stores relationships in a temporary array. The infinite list of related records will not auto-update to reflect if / when a new relationship is added on that page.
- This addon will attempt to fetch every record. In a large recordset, this may lead to severe UX and rendering issues. *Use this very, very sparingly*.

## Use case
Pending backend changes to support instant search, this will be required for front-end filter functionality.

## Goals
This PR, fully realized, should demonstrate proof of concepts for:
- [x] Fetch all record functionality for a record accessed directly: `store.query('node')`
- [x] Fetch all record functionality for relationships: `oneUser.nodes`
   - Plan: allow customizable query function (model.query), and otherwise keep interface similar, using `model.relname` as the passed-in infinitymodel. The major challenge will be to make the relation available as a separate item so the infinitymodel can update it- can we use a straight-up ember array as an infinitymodel??
- [x] Route mixins to handle fetching conveniently (along with dire disclaimers)

## Routes Added/Updated
Proof of concept routes:
- `/usernodes`- for `user.nodes` relationship
- `/allnodes`- for a direct model query


addon/serializers/osf-serializer.js
```diff
@@ -99,6 +99,7 @@ export default DS.JSONAPISerializer.extend({
         let documentHash = this._super(...arguments);
         documentHash.meta = documentHash.meta || {};
         documentHash.meta.pagination = Ember.get(payload || {}, 'links.meta');
+        documentHash.meta.total = Math.ceil(documentHash.meta.pagination.total / documentHash.meta.pagination.per_page);
```
samchrisinger: Shouldn't this just be total?

abought: APIv2 returns total number of *records*, but ember infinity relies on being told the total number of *pages*. 

I can tweak field name and mixin settings to make the name of this field more explicit, if that would help.

samchrisinger: Thanks for the explanation. I think it's fine as is.

addon/mixins/infinity-custom.js
```diff
@@ -0,0 +1,85 @@
...
+     @param {String} modelName The name of the model.
+     @param {Object} options Optional, the perPage and startingPage to load from.
+     @param {Object} boundParams Optional, any route properties to be included as additional params.
+     @return {Ember.RSVP.Promise}
+     */
+    infinityModel(modelName, options, boundParams) {
+        this.set('_infinityModelName', modelName);
+
+        options = options ? assign({}, options) : {};
+        const startingPage = options.startingPage === undefined ? 0 : options.startingPage - 1;
```
samchrisinger: Would this be better off as a computed?

samchrisinger:
Biggest concern for me is the difficulty following the code in the ember-infinity subclasses (mostly due to calls of methods defined elsewhere) ... but I don't really have any constructive advice on this. All in all, lgtm.

abought:
> Biggest concern for me is the difficulty following the code in the ember-infinity subclasses (mostly due to calls of methods defined elsewhere) ... but I don't really have any constructive advice on this. All in all, lgtm.

Yeah- for both good and ill, this really is tightly coupled to the underlying addon. (especially the need to call `infinityModel` instead of `store.query`). 

We could alias that to a wrapper function ("`fetchAllQuery`") to make the "internals" more visible for people implementing this addon. It's kind of a blunt instrument as written.

samchrisinger:
@abought if it's ok with you, I'd like to go ahead and merge. We'll add tests and iterate as needed.

addon/mixins/infinity-custom.js
abought: Yeah. This while method would be better off not needing to exist... it's mostly a cut-and-paste tweak to the parent method. 

The main thing this custom class does is to make the store query method configurable (inserted in the middle of the function), so we can query relationships. (which don't go through `this.store.query`)

In the next iteration, I may take another crack at consolidating this duplicate code.

abought:
> @abought if it's ok with you, I'd like to go ahead and merge. We'll add tests and iterate as needed.

Sounds ok to me. I'm more than happy to make improvements as desired; at times writing this, it felt like the software equivalent of a sledgehammer duct-taped to an edsel.

PR #63, opened 2016-06-10
File detail page and readonly file browser [EOSF-55] [EOSF-57]

aaxelb:
## Ticket
[EOSF-55](https://openscience.atlassian.net/browse/EOSF-55) (read-only file browser and file detail page) and [EOSF-57](https://openscience.atlassian.net/browse/EOSF-57) (file checkout)
(Sorry for the poor JIRA etiquette)

Also lays some groundwork for [EOSF-56](https://openscience.atlassian.net/browse/EOSF-56) and [EOSF-58](https://openscience.atlassian.net/browse/EOSF-58).
# Purpose
The goal for this PR/ticket is a file detail page that looks and acts like that on OSF, including a file tree in the left sidebar.

# Notes for Reviewers
Most of the TODOs aren't blockers for this PR, just indicators of future work toward feature parity with fangorn/treebeard. The `file-browser` code (and related components) will mutate as we add features and likely move toward a flat, non-hierarchical file display.

Test the actions at the top of the file detail page, especially checkout/checkin, which required a hacky file serializer special case.

## Routes Added/Updated
`/nodes/detail/files/provider/file` updated to look/act more like the OSF file detail page:
screen shot 2016-06-15 at 11 20 37 am

`/nodes/detail/files/provider/file/revisions` added to display and allow downloading older versions of a file:
screen shot 2016-06-15 at 11 21 26 am

PR #71, opened 2016-06-16
Flat-view file browser [EOSF-68]

aaxelb:
## Ticket
https://openscience.atlassian.net/browse/EOSF-68
# Purpose
Replace the tree-view file browser in #63 with a flat-view (à la Dropbox), which will be simpler for the user (especially on mobile) and let us add CRUD operations and drag/drop without adding much complexity.

Right now it's a functional file browser, but there are a couple TODOs remaining (e.g. the list won't update if files are added or changed, and there's limited error handling) and it could use some UI design. 
# Notes for Reviewers
Test that the breadcrumbs are always accurate as you browse around. Clicking on file/node links should open their corresponding detail pages, and double clicking on a node's row opens it like a folder.

It should work for long lists of files/nodes, adding them one page at a time as they load. There's another (partial) approach to relationship pagination in [another branch](https://github.com/aaxelb/ember-osf/blob/feature/paginated-relationship/addon/mixins/paginated-has-many.js), if anyone's curious. I tried using that branch's approach in this one, but ran out of time and reverted to a more stable state before handing this off.


pattisdr:
I really like this flat view - navigation in the breadcrumbs seems to work well.

Issues to investigate - 
Functionality:
- [ ] When you click on a file name and it renders on the right, the entire document tree collapses.  If I had navigated to a deeply nested document, I might want to view the other items in the folder, not have to start from scratch.
- [ ] Only show buttons if they are allowed to be clicked.  For example, I think you can only check out documents if they are on OSFStorage.  I don't think you can check out something stored on Github for example, so the checkout button shouldn't be visible for these other providers.
- [ ] When the file tree loads, you initially have to double click on a provider to open it, and once this has happened once, you only have to single click on providers.

UI
- [ ] Long file names loop to the next line and overlap data there.
- [ ] A spinner needed when the MFR rendering is loading
- [ ] More vertical spacing between deeply nested breadcrumbs
- [ ] Mobile view - potentially large whitespace in between file tree and file rendering.

Error messages - 
- [ ] When user navigates to file that has been deleted

PR #75, opened 2016-06-20
NodeActionsMixin

samchrisinger:
replaces https://github.com/CenterForOpenScience/ember-osf/pull/66

# Purpose

Moving actions and action listeners to callbacks promotes reusability and shareability of code. This allows for a single canonical understanding of how to interact with the ember-osf models. 

This PR attempts to consolidate some of the Node actions into a single Mixin.

# Testing notes

Behaviors touched:

- Add/remove institution
- Add child
- Update contributors
- Add/remove contributor


pattisdr:
Starting review now.

addon/mixins/node-actions.js
```diff
@@ -0,0 +1,100 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+    node: null,
+    model: null,
+    _node: Ember.computed.or('node', 'model'),
```
pattisdr: Can you explain to me how this is getting set equal to the node?

samchrisinger: This is shorthand for:

```javascript
Ember.computed('node', 'model', function() {
    return this.get('node') || this.get('model');
})
```

pattisdr: I just don't understand where 'node' is getting set in the first place.   On the node detail route, I see the `return this.store.findRecord('node', params.node_id);` - because it is looking up an individual node, is that how it knows what this.get('node') is without having to explicitly define 'node'?

addon/mixins/node-actions.js
```diff
@@ -0,0 +1,100 @@
...
+            if (category) {
+                node.set('category', category);
+            }
+            if (description) {
+                node.set('description', description);
+            }
+            if (isPublic !== null) {
+                node.set('public', isPublic);
+            }
+            return node.save();
```
pattisdr: Why is it necessary to do `return node.save()` in all these actions?  Is it so the promise is returned?

addon/mixins/node-actions.js
samchrisinger: Since this is mixin class, it's expected to be combined with a route, controller, or component. For:
- a route: falls back to model, or uses `Ember.computed.alias` to point to another attribute, e.g.:
```javascript
model() {
    node: Ember.computed.alias('model.node'),
    return new Ember.RSVP.hash({
        node: ...,
        user: ...
    })
}
```
- a controller: node could be set by `setupController` in the route or could fall back to model
- a component: node is passed down from the parent component/controller

addon/mixins/node-actions.js
```diff
@@ -0,0 +1,100 @@
...
+        },
+        deleteNode() {
+            var node = this.get('_node');
+            return node.destroyRecord();
```
pattisdr: This is a good example of an action that should have a transition after the promise is returned.  Did you remove this because we don't want to enforce in this mixin where to redirect?

addon/mixins/node-actions.js
samchrisinger: Yes- exactly, which allows for responding the to the save result (e.g. https://github.com/CenterForOpenScience/ember-osf/pull/75/files#diff-2df153f5476f3656ebfb69d48424f742R15)

addon/mixins/node-actions.js
samchrisinger: Yes. I feel that a mixin shouldn't make any assumptions about the applications routing (there may be no such thing as a 'node detail' route).

addon/models/node.js
```diff
@@ -50,7 +52,12 @@ export default OsfModel.extend({
     comments: DS.hasMany('comments'),
     contributors: DS.hasMany('contributors', {
         updateRequest: {
-            requestType: (snapshot, relationship) => snapshot.hasMany(relationship).filter(each => Object.keys(each.changedAttributes()).indexOf('userId') !== -1).length > 0 ? 'POST' : 'PATCH',
+            requestType: (snapshot, relationship) => {
+                if (snapshot.hasMany(relationship).filter(each => each.record.get('isNew')).length) {
+                    return 'POST';
+                }
+                return 'PATCH';
+            },
```
pattisdr: This is much more readable!

tests/dummy/app/components/contrib-manager/template.hbs
```diff
@@ -0,0 +1,54 @@
...
+    {{#each contributors as |contrib|}}
+        <tr>
+            <td>
+                {{contrib.id}}
+            </td>
+            <td>
+                {{if contrib.unregisteredContributor contrib.unregisteredContributor contrib.users.fullName}}
+            </td>
+            <td>
+                <select onchange={{action (action 'permissionChange' contrib) value='target.value'}} >
```
pattisdr: Can you explain why these nested actions are here?

tests/dummy/app/templates/nodes/detail/index.hbs
```diff
@@ -151,14 +102,11 @@
...
     <div>
         <h3> First Component </h3>
-        <label>Title:</label>
-        {{input type="text" value=child1 placeholder="Component's title" required=true}}
+        <label>Title:</label> {{input type="text" value=child1 placeholder="Component's title" required=true}}
         <h3> Second Component </h3>
-        <label>Title:</label>
-        {{input type="text" value=child2 placeholder="Component's title" required=true}}
+        <label>Title:</label> {{input type="text" value=child2 placeholder="Component's title" required=true}}
         <p><button {{action "addChildren" child1 child2 }}>Create</button></p>
```
pattisdr: Nothing handling this action now but you might have intended it to be in a separate PR.

tests/dummy/app/components/contrib-manager/component.js
```diff
@@ -0,0 +1,34 @@
...
+            this.sendAction('removeContributor', contrib);
+        },
+        permissionChange(contributor, permission) {
+            this.set(`permissionChanges.${contributor.id}`, permission.toLowerCase());
+        },
+        bibliographicChange(contributor, isBibliographic) {
+            this.set(`bibliographicChanges.${contributor.id}`, isBibliographic);
+        },
+        updateContributors() {
+            this.attrs.editContributors(
```
pattisdr: Why do you need to do this.attrs() instead of this.sendAction() here?

tests/dummy/app/controllers/nodes/detail/index.js
```diff
@@ -1,25 +1,23 @@
...
-            // Adds updated bibliographic info for a certain contributor
-            var bibliographic = target.checked;
-            var contributorId = target.value;
-            this.editedBibliographic[contributorId] = bibliographic;
+        updateNode() {
+            this.set('isSaving', true);
+            return this._super(...arguments)
+                .then(() => {
+                    this.set('isSaving', false);
+                    this.get('toast').success('Node updated successfully');
```
pattisdr: I like this!

tests/dummy/app/templates/nodes/index.hbs
```diff
@@ -18,7 +18,7 @@
     {{input type="text" value=title placeholder="Node's title" required=true}}
     <label>Description (optional):</label>
     {{input type="text" value=description placeholder="Node's description"}}
-    <p><button {{action "createNode" title description}}>Create Node</button></p>
+    <p><button onclick={{action 'createNode' title description}}>Create Node</button></p>
```
pattisdr: What is the reason for specifying onclick here?

tests/dummy/app/components/contrib-manager/template.hbs
samchrisinger: There is some explanation here: http://emberjs.com/api/classes/Ember.Templates.helpers.html#toc_invoking-an-action, but please let me know if you have questions after reading this.

tests/dummy/app/templates/nodes/index.hbs
samchrisinger: It's really not necessary. I kind of prefer the explicit `onclick` just because <button> is native html, but this could be removed.

tests/dummy/app/components/contrib-manager/component.js
samchrisinger: `sendAction` doesn't return a value (https://github.com/emberjs/ember.js/blob/v2.6.0/packages/ember-views/lib/components/component.js#L283), whereas calling the method directly from attrs allows us to respond to create actions that return values.

pattisdr:
Maybe on the API side we could add more logic for bulk updating contributors.  Doing bulk update wraps everything in one transaction so it can get rolled back, but there's no guarantee that the actions happen in the correct order.  So if you do a few bulk requests trying to toggle between which contributor is bibliographic for example, you'll eventually get a 400.

Or I am getting errors that I am the only admin, when I am only changing bibliographic info.  This bulk update seems unreliable right now.

tests/dummy/app/components/contrib-manager/component.js
pattisdr: Where do you use the returned value?

tests/dummy/app/templates/nodes/index.hbs
pattisdr: It's fine! Just curious.

samchrisinger:
> This bulk update seems unreliable right now.

I agree 100%. This PR: https://github.com/CenterForOpenScience/ember-osf/pull/78 makes big steps towards better bulk support.

tests/dummy/app/components/contrib-manager/component.js
samchrisinger: Here it's unused. I should probably just call `sendAction`

pattisdr:
Pass finished - good work @samchrisinger! Only requested change here https://github.com/CenterForOpenScience/ember-osf/pull/75#discussion_r67944906.

This PR doesn't change a lot of functionality - there's just a lot better organization here.  Moves a lot of the node actions into a mixin and makes them more generalizable by not specifying redirects or permission checks. Making the contributor management a component is a big plus too.  Also, I saw a lot of needed template clean-up.

tests/dummy/app/templates/nodes/detail/index.hbs
samchrisinger: Good call. I'm feeling that we should strongly discourage bulk creates, and will remove this piece of UI.

addon/mixins/node-actions.js
pattisdr: Ah okay, thanks :)

tests/dummy/app/components/contrib-manager/template.hbs
pattisdr: hmm I need to read more about this.  "currying" is new to me.

package.json
```diff
@@ -45,7 +46,9 @@
     "ember-faker": "^1.1.0",
     "ember-infinity": "0.2.3",
     "ember-load-initializers": "0.5.1",
+    "ember-moment": "6.1.0",
```
abought: This is already required on line 71; is the redundancy necessary?

tests/dummy/app/components/contrib-manager/template.hbs
pattisdr: Are you using this so you can pass in both the contributor and the target.value?

samchrisinger: Right. As far as I understand, the `value` keyword always re-writes the first argument. At the same time, having a non-keyword argument (`contrib`) after a keyword argument (`value=`) is invalid syntax. So the only way I can tell to get a reference to the contributor is to use the approach.This could be translated to:

```javascript
var innerAction = permissionChange.bind(null, contrib);
select.onchange = function(target) {
    innerAction(target.value)   
}
```

package.json
samchrisinger: Nope, good catch!

tests/dummy/app/components/contrib-manager/template.hbs
pattisdr: This is good to know!  When I wrote contributors the first time, I was trying to pass in the arguments as contribID:permissions, and then split it up at the colon later.  I knew it wasn't a great way to do it, but didn't know how else to get what I needed.

pattisdr:
Changes look good to me.  I really like this mixin - it makes things a lot simpler.

PR #76, opened 2016-06-21
[WIP] Unit tests for fetch-all mixin

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-70

# Purpose
Add unit tests for fetch-all mixin.

Also adds `NOTICE.md` to track licensing obligations.

# Notes for Reviewers
I had a bit of trouble getting sinon spies to recognize the module / beforeEach code, so ended up using manual mocks. Credit to https://github.com/hhff/ember-infinity/blob/master/tests/unit/mixins/route-test.js for inspiration.

- Test that fetch-all function can work for a normal model directly
  - Results are put in the specified place, and multiple pages of results are automatically fetched (not just one page)
- Test that relationships fetching happens as expected
  - The presence of a relationship name triggers relationship fetching
  - Results are put in a pre-defined place
  - Special behaviors (like dealing with page 1 of results) are handled automatically when this mixin is in "relationship fetching mode"

## Routes Added/Updated
None. Added helper method to mixin to facilitate testing.


package.json
```diff
@@ -46,8 +46,6 @@
     "ember-infinity": "0.2.3",
     "ember-load-initializers": "0.5.1",
     "ember-resolver": "2.0.3",
```
abought: I removed these lines after realizing that my initial experiments didn't use sinon after all... but it's not clear that anyone else's tests require these dependencies either. Should I keep these lines in for now?

samchrisinger: Probably something I added and never removed, seems fine to me.

samchrisinger:
lgtm, thanks!

PR #77, opened 2016-06-21
feature/Clear form after tag added [EOSF-73]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-73

# Purpose

Intern request. Clears form after tag added.  Based off of patterns in @abought's commenting widget https://github.com/CenterForOpenScience/ember-osf/pull/61.

### DIFF
--- a/addon/mixins/taggable-mixin.js
+++ b/addon/mixins/taggable-mixin.js
@@ -8,7 +8,7 @@ export default Ember.Mixin.create({
             Ember.A(currentTags);
             currentTags.pushObject(tag);
             resource.set('tags', currentTags);
-            resource.save();
+            return resource.save()
         },
         removeATag(tag) {
             var resource = this.get('model');
--- a/addon/mixins/taggable-mixin.js
+++ b/addon/mixins/taggable-mixin.js
@@ -8,7 +8,7 @@ export default Ember.Mixin.create({
             Ember.A(currentTags);
             currentTags.pushObject(tag);
             resource.set('tags', currentTags);
-            return resource.save()
+            return resource.save();
         },
         removeATag(tag) {
             var resource = this.get('model');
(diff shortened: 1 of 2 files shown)


addon/mixins/taggable-mixin.js
```diff
@@ -8,7 +8,7 @@ export default Ember.Mixin.create({
             Ember.A(currentTags);
```
abought: The return value is unused; what is this line doing?

pattisdr: When I wrote this, I was having a hard time getting changedAttributes to update. I first did `file.get('tags').pushObject(tag)`, but it was saying that pushObject was not a function, so I changed it to Ember.A(file.get('tags')).pushObject(tag). But this results in changedAttributes being an empty object. So no PATCH request was sent.

The only way I could get changedAttributes to modify was to copy the current tags (not referencing original) and push new tag onto list, and then set tags equal to this list.  

Sam said that Henrique's work may fix this needing a roundabout way to modify the tags.

abought: Right, but I think you've made two copies, and this second one appears unused. Does it work without this line?

(both `slice` and `Ember.Array` should return a new copy)

pattisdr: It does not work without this line- it says that `currentTags.pushObject is not a function`.  I was using slice to make a copy and Ember.A to turn it into something I could do pushObject on.

pattisdr: But you're right, I could simplify with `var currentTags = Ember.A(resource.get('tags'));`

abought: Hmm. It appears that `Ember.A` really does work this way- weird, but interesting! Thanks.

Page suggesting it modifies the original in place: https://guides.emberjs.com/v2.3.0/configuring-ember/disabling-prototype-extensions/#toc_arrays

Page suggesting that it totally doesn't modify the original object:
http://emberjs.com/api/#method_A

pattisdr: I think I still need the slice or something similar - I need to make a copy that doesn't reference the original in order for the changes to persist.

PR #78, opened 2016-06-21
[RFR] Improved relationship dirty tracking

samchrisinger:
Depends on https://github.com/CenterForOpenScience/ember-osf/pull/75

Ticket: https://openscience.atlassian.net/browse/EOSF-79

# Purpose

The old way we were tracking dirty relationships had some issues. This moves all of the dirty checking into the osf-models save method (lazy evaluation of dirtiness). Additionally it does a better job of notifying the store when relationships have changed-- since our dirtyTracking depends on accurately knowing the 'canonical' (last clean) state of a record, this is an essential improvement. 

# Testing Notes

It would be good to as a minimum check that actions specified in https://github.com/CenterForOpenScience/ember-osf/pull/75 still work.


addon/adapters/osf-adapter.js
```diff
@@ -18,97 +24,199 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+    _buildRelationshipURL(snapshot, relationship) {
         var links = relationship ? snapshot.record.get(
             `relationshipLinks.${Ember.String.underscore(relationship)}.links`
         ) : false;
         if (links && (links.self || links.related)) {
             return links.self ? links.self.href : links.related.href;
+        }
+        return null;
+    },
+    _createRelated(store, snapshot, createdSnapshots, relationship, url, isBulk = false) { // jshint ignore:line
```
abought: As a totally general remark, I'm encouraging everyone (including myself) to add YUIDoc comments to new functions, documenting their intended purpose and arguments (or at least any unique ones).

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +24,199 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+            });
         }
-        return serialized;
+        return this.ajax(url, requestMethod, {
+            data: data,
+            isBulk: isBulk
+        }).then(res => {
+            if (!res) {
+                if (requestMethod === 'DELETE') {
+                    res = [];
```
abought: Possibly a question about ember internals, but... why do we set `res` to an empty list for deletes? (but not other cases)

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +24,199 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
-                url: this._buildRelationshipURL(type.modelName, snapshot.id, snapshot, 'updateRecord', query, relationship),
-                serialized: this._serializeHasMany
-            },
-            relationMeta.options.updateRequest
-        );
-        return this.ajax(options.url, options.requestType(snapshot, relationship), {
-            data: options.serialized(serialized),
-            isBulk: options.isBulk(serialized)
-        }).then(() => snapshot.record.clearDirtyRelationship(relationMeta));
+    _handleRelatedRequest(store, type, snapshot, relationship, change) {
```
abought: Some documentation (especially of the change parameter and allowed values) would be extremely helpful for maintainability here. :)

addon/models/node.js
```diff
@@ -40,52 +36,21 @@ export default OsfModel.extend({
         inverse: 'children'
     }),
     children: DS.hasMany('nodes', {
-        inverse: 'parent',
-        updateRequest: {
-            requestType: () => 'POST'
```
abought: For my own referebce: This relationship meta config field was required to make comments work. Will the removal affect the new PR as well?

addon/adapters/osf-adapter.js
samchrisinger: The 204 empty response from our API confuses Ember. Representing this as empty list instead is an easy workaround.

addon/models/node.js
```diff
@@ -105,5 +70,25 @@ export default OsfModel.extend({
...
+                contributors.members.list.filter(m => !m.record.get('isNew') && Object.keys(m.record.changedAttributes()).length > 0)
+            );
+            // Contributors are a 'real' delete, not just a de-reference
+            this.set(
+                '_dirtyRelationships.contributors.delete',
+                this.get('_dirtyRelationships.contributors.remove')
+            );
+            this.set('_dirtyRelationships.contributors.remove', []);
+        }
+        return promise;
```
abought: It seems like this function does two things: sending a request for the record, and also managing its relationships.

Is there a situation where one of those changes could fail, but not the other? Should we take that into account to ensure the state of data stays in sync?

For example: it seems we call `this.set` for everything, regardless of whether the promise succeeds.

tests/dummy/app/app.js
```diff
@@ -6,6 +6,7 @@ import config from './config/environment';
 let App;
 
 Ember.MODEL_FACTORY_INJECTIONS = true;
+Ember.run.backburner.DEBUG = true;
```
abought: This setting might be helpful, but the docs indicate that it slows down even normal development substantially. Would you be ok with keeping it turned off for the dummy app, unless explicitly needed?

https://guides.emberjs.com/v2.4.0/configuring-ember/debugging/#toc_errors-within-code-ember-run-later-code-a-href-https-github-com-ebryn-backburner-js-backburner-js-a

addon/models/node.js
samchrisinger: Part of the goal of this PR is to remove this.

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +24,199 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+    },
+    _removeRelated(store, snapshot, removedSnapshots, relationship, url, isBulk = false) {
+        return this._doRelatedRequest(store, snapshot, removedSnapshots, relationship, url, 'DELETE', isBulk).then(response => response || []);
+    },
+    _deleteRelated(store, snapshot, removedSnapshots) { // jshint ignore:line
+        return this._removeRelated(...arguments).then(() => {
+            if (removedSnapshots.record) {
+                removedSnapshots = [removedSnapshots];
+            }
+            removedSnapshots.forEach(s => s.record.unloadRecord());
```
abought: Previously we'd avoided direct store manipulation. To help me read this code, can you please give me an example of the intended behavior that this is designed to improve upon?

How are "remove" and "delete" intended to be different?

abought:
Thanks for the testing notes. To help me know how rigorous to be: have you been able to confirm these actions work with the code in its present state? (after refactorings) 

Are there any actions that require special attention? (likely edge cases)

addon/adapters/osf-adapter.js
samchrisinger: Remove: I remove an institutional reference, but that record should not get unloaded from the store
Delete: I delete a contributor or a child node, and that record should be removed from the store. 

Since we have to manage the relationship requests manually, we also have to unload the record-- this is part of the public API for DS.Model.

abought:
Purely for future reference:

Ember-data has a PR in progress that will allow rollback of dirty relationship state. We'll want to keep an eye on that to ensure future compatibility of our dirty state tracking with the planned rollback mechanism. https://github.com/emberjs/data/pull/3698

addon/adapters/osf-adapter.js
abought: Hmm. Ultimately, I assume that part of the distinction is because we have separate models for contributors vs users?

I can see this making sense if, eg the user still exists, but the contributor reference was tied to a specific node.

(since we're not deleting the user from the backend, just the relationship)

samchrisinger:
I'm happy to delete lots of code when ember-data adds support for tracking relationship dirty states, but I was pretty disappointed with the number of seemingly stagnant PRs adding this that I saw.

abought:
When I try to affiliate a project with an institution it is already connected to:
- It doesn't send a post request (which is good), but
- It does send another get request to https://staging-api.osf.io/v2/institutions/cos/

Is the extra network request necessary? It's not a blocker for merge right now due to current issues... but it may be a sign of some logic firing when it shouldn't.

samchrisinger:
Yes. The code here always tries to fetch the Institution for the given id before adding it

abought:
On the "node detail" page, "contributors section":
Sample URL  http://localhost:4200/nodes/m9pv6

When I update an existing contributor (r+w --> read permissions), then click "save changes", the table disappears.

It returns when I refresh the page:

screen shot 2016-06-22 at 3 11 10 pm

screen shot 2016-06-22 at 3 10 19 pm

abought:
Pass complete. Two categories of comments:
- Some of this is tricky to review because it depends on subtle ember internals. That's not surprising, and it'll be resolved by careful notes and good unit tests.

- There are a few behavioral oddities that I'd consider blockers for merging:
  - [ ] Relationships fields may be marked for update regardless of whether the promise succeeds. Are we trusting the `osf-model` save mechanisms to clean this up correctly? https://github.com/CenterForOpenScience/ember-osf/pull/78#discussion_r68097164
  - [ ] As discussed on flowdock, I see the following error when loading a sample project page: `ember.debug.js:32116 Error: Assertion Failed: Encountered a resource object with an undefined type (resolved resource using dummy@serializer:node-link:)`
  - [ ] The contributors section is showing weirdness. It's possible this is tied to how we manually manage the store. https://github.com/CenterForOpenScience/ember-osf/pull/78#issuecomment-227847360

addon/models/node.js
samchrisinger: > Is there a situation where one of those changes could fail, but not the other? 

This is at least partially taken into account. If you [look at the internals of _setupRelationships](https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L2455), it only updates the fields for which data is included. So the strategy here is to be conservative about what we tell the store has updated.

 One important note: if the inverse model is dirty, `OsfModel#save` _always_ returns the response from the super call. Looking more at the Store's internals:
See 
- https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L1646
- https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L2430
This should be a reliable way to communicate state changes. 

Admittedly this is imperfect and will require future improvements.

samchrisinger: > it seems we call this.set for everything, regardless of whether the promise succeeds

True. But the _dirtyRelationships collection is (designed to be 😁 ) one-to-one, so _assuming the canonicalStates are updated correctly_, this is a non-issue.

samchrisinger:
@abought can you please take another pass of this?

abought:
Will take a look this evening- nice to see this!

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +20,256 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+     * @param {String} relationship
+     * @param {String} url
+     * @param {Boolean} isBulk
+     **/
+    _createRelated(store, snapshot, createdSnapshots, relationship, url, isBulk = false) { // jshint ignore:line
+        // TODO support bulk create?
+        // if (isBulk) {
+        //
+        // }
+        return createdSnapshots.map(s => s.record.save({
```
abought: In certain circumstances, I receive an error:

> vendor.js:165498 Uncaught TypeError: createdSnapshots.map is not a function

Example: trying to add `z4esu` as a bibliographic read contributor from the node detail page (user management panel) after a refresh. If I try doing it again on the page, I get:
> vendor.js:30475 Uncaught Error: Assertion Failed: The id z4esu has already been used with another record of type dummy@model:contributor:.

The user is not added to the list (either visually, or after refreshing the page). It does not appear that a network request is sent.

This is the test project we've used before, so you should have access.

pattisdr: I am also seeing this, but in every circumstance that this method is run.  Trying to map the snapshot is not working.

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +20,256 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+                data,
+                store.modelFor(type),
+                relatedSnapshots, {
+                    forRelationship: true,
+                    isBulk: isBulk
+                }
+            );
         } else {
-            serialized = serialized[0];
+            data.data = relatedSnapshots.map(relatedSnapshot => {
```
pattisdr: Also getting same JS error here in trying to map the snapshot. `Uncaught TypeError: relatedSnapshots.map is not a function`

addon/adapters/osf-adapter.js
abought: Confirmed that this error case has been resolved in #88 for me. Works after adding, removing, and refresh. 

The assertion still happens, but I think that's by design and it prevents bad behavior, so totally ok with that.

PR #80, opened 2016-06-22
Feature/Registration Mixin [EOSF-77]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-77

# Purpose

Consolidate registration actions in a single mixin similar to node actions mixin https://github.com/CenterForOpenScience/ember-osf/pull/75.

Also, makes some requested changes to Draft Registration Model PR https://github.com/CenterForOpenScience/ember-osf/pull/69

# Notes for Reviewers

- Adds some handling for the Veer Preregistration schema.  More thought will be needed on how to best handle all of the different types of schemas.
- Building the request for editing a draft's metadata even more messy after adding handling of Veer schema. Will need refactoring at some point.
- If a draft has been created on the OSF, and then is edited in the dummy app, it may not work.  Metadata on the OSF has blank strings stored as field answers, but API doesn't accept sending back an empty string as a multiple choice response, for example.  So API request will need to be modified to strip out these fields that are not actually answers to questions. Right now I am just adding the changes to the current metadata and sending it back.  (If the draft was created in Ember-OSF though it works fine).


addon/mixins/registration-actions.js
```diff
@@ -0,0 +1,68 @@
+
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+    node: null,
+    draft: null,
+    model: null,
+    _node: Ember.computed.or('node'),
```
samchrisinger: Using `_node` is probably unneeded if there is a just one attribute for the node.

addon/mixins/registration-actions.js
```diff
@@ -0,0 +1,68 @@
...
+    _draft: Ember.computed.or('draft', 'model'),
+    /** Updates current registration metadata with new responses to questions.
+     **/
+    _updateMetadata(d, u) {
```
samchrisinger: Please avoid single-character variable names (except _maybe_ in single-line anonymous functions)

tests/dummy/app/routes/nodes/detail/draft-registrations/detail.js
```diff
@@ -1,72 +1,14 @@
...
 
 export default Ember.Route.extend({
     model(params) {
         let node = this.modelFor('nodes.detail');
-        let draft = this.store.peekRecord('draft-registration', params.draft_registration_id);
-        return Ember.RSVP.hash({
-            node: node,
-            draft: draft
-        });
+        this.store.adapterFor('draft-registration').set('namespace', 'v2/nodes/' + node.id);
```
samchrisinger: If we need special logic in the draft-registration adapter we should do this on the class rather than re-setting the namespace as needed.

pattisdr: What do you mean by do it on the class?

samchrisinger: If the `DraftRegistrationAdapter` needs some custom behavior as a subclass of the `ApplicationAdapter`, then that logic should be implemented there. Is there a relationship on the Node model we can use for this?

pattisdr: I can see why there are better places to put this but I don't know where it should go. My reasoning for putting it here is that I needed access to the node. Maybe we need to think bigger picture about this - findRecord is not really setup to locate nested resources. It's fine most of the time, like logs, for example, you can access nodes/<node_id>/logs/ but if you want log detail, you go to logs/log_id.

This doesn't work for drafts because there is no drafts/draft_id endpoint.

PR #83, opened 2016-06-24
Add file-widget :rainbow:

samchrisinger:
🌟


index.js
```diff
@@ -26,7 +26,8 @@ module.exports = {
             clientId: SETTINGS.CLIENT_ID,
             scope: SETTINGS.OAUTH_SCOPES,
             apiNamespace: 'v2', // URL suffix (after host)
-            backend: BACKEND
+            backend: BACKEND,
+            redirectUri: SETTINGS.REDIRECT_URI
```
aaxelb: Should `REDIRECT_URI` be added to the `_local.yml` blueprint?

samchrisinger: Good call, thanks for the catch!

PR #85, opened 2016-06-24
Feature/Add Search to Navbar [EOSF-26]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-26

# Purpose

Adds search to navbar.  Gets rid of unnecessary imports.

# Notes for Reviewers

### DIFF
--- a/addon/components/search-dropdown/template.hbs
+++ b/addon/components/search-dropdown/template.hbs
@@ -2,15 +2,14 @@
     <div class="container">
         <div class="row">
             <div class="col-md-12">
-                <form class="input-group" data-bind="submit: submit">
-                    <input id="searchPageFullBar" name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search" data-bind="value: query, hasFocus: true">
+                <form class="input-group" {{action "search" on="submit"}}>
+                    <input id="searchPageFullBar" name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search">
                     <label id="searchBarLabel" class="search-label-placeholder" for="search-placeholder">Search</label>
                     <span class="input-group-btn">
-                        <button type=button class="btn osf-search-btn" data-bind="click: submit"><i class="fa fa-circle-arrow-right fa-lg"></i></button>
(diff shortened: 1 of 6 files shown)


addon/components/search-dropdown/template.hbs
```diff
@@ -1,16 +1,15 @@
-<div class="osf-search" data-bind="fadeVisible: showSearch" style="display: none">
+<div class="osf-search">
     <div class="container">
         <div class="row">
             <div class="col-md-12">
-                <form class="input-group" data-bind="submit: submit">
-                    <input id="searchPageFullBar" name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search" data-bind="value: query, hasFocus: true">
+                <form class="input-group" {{action "search" on="submit"}}>
+                    <input id="searchPageFullBar" onchange={{action "updateQuery"}} name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search">
```
samchrisinger: You can use `value="target.value` here, and then update the signature of `updateQuery` to just take a value

samchrisinger:
One tiny comment, otherwise lgtm

addon/components/search-dropdown/template.hbs
pattisdr: I tried that originally, but then the placeholder was target.value, which I didn't want.

samchrisinger:
@pattisdr looks like the search bar might be getting hidden for small screens. Can you please take a look at this?

pattisdr:
Yes - thanks, I didn't check small screens.

addon/components/search-dropdown/template.hbs
samchrisinger: I'm not seeing this behavior. When I make the change locally the "Search' placeholder text works as expected and I get the correct value in `updateSearch`. Can you please double check this?

pattisdr: I must be misunderstanding what you want here. 

I did this:
```         
 <input id="searchPageFullBar" value=target.value onchange={{action "updateQuery"}} name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search">
````
which puts placeholder as target.value.  How do I reference this value to pass it into updateQuery?

pattisdr:
Requested changes made.

PR #88, opened 2016-06-28
Feature/Dirty Relationship Tracking Polishing [EOSF-79]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-79

# Purpose

Wraps up https://github.com/CenterForOpenScience/ember-osf/pull/78.  

Have looked at changes to osf-adapter, still need to understand changes to dirtyRelationships.

# Notes for Reviewers

❗ The main fix was making multiple requests without refresh was causing errors in the console because dirty relationships weren't being cleared.  To test I would make multiple types of requests that add relationships one after another.

My changes:
 
- fetched metaschemas instead of hardcoding
- fixed js errors related to mapping snapshots
- modified _doRelatedRequest - sometimes was getting request with list in list, maybe due to fix for previous item
- Handling for adding a project to a collection (first attempt).  Linked_node serializer added because request to add a  linked_node to a collection needs to look like adding an institution to a node.  But there's no linked_node model, linked nodes are actually nodes. May need to rethink this.
- Changes for running addCanonicalRecord or removeCanonicalRecord - wasn't being cleared.
- Passing in snapshots instead of internalRecords to canonicalRecords

Issues:
- Deleting node links and deleting drafts don't properly update UI.  
- Bulk updates to contributors still unreliable (maybe need changes API side?)


tests/dummy/app/routes/nodes/detail/draft-registrations/index.js
```diff
@@ -8,6 +8,8 @@ export default Ember.Route.extend({
     },
     setupController(controller, model) {
         this._super(controller, model);
+        var metaschemas = this.store.findAll('metaschema');
+        controller.set('metaschemas', metaschemas);
```
abought: Are there any concerns about, eg, needing to wait on the resolution of the promise?

(certainly for a computed; I'm less sure about what happens you you set a promise as a property on the controller directly)

I couldn't examine this locally,  so instead I'll just ask you. I received an error response with a 200 status code from the API for a project with 1 draft and one completed registration. The API issue is unrelated to this PR. (pinged @briang for varnish triage)
```
Request URL:https://staging-api.osf.io/v2/nodes/8k3w2/draft_registrations/
Request Method:GET
Status Code:200 OK

{"data":[{"errors": [{"detail": "Not found."}]}],"links":{"first":null,"last":null,"prev":null,"next":null,"meta":{"total":1,"per_page":10}}}
```

addon/serializers/linked-node.js
```diff
@@ -0,0 +1,14 @@
+import OsfSerializer from './osf-serializer';
+
+export default OsfSerializer.extend({
+    serializeIntoHash(hash, typeClass, snapshot, options) {
+        if (options.forRelationship) {
+            hash.data = [{
```
abought: Does this override any preexisting `data` property? Do we mean to add to the existing properties instead?

tests/dummy/app/routes/nodes/detail/draft-registrations/index.js
pattisdr: Local is working fine for me - I have all the metaschemas loaded when I navigate to the draft index view.  Making sure the promise has resolved might be important however.  Do you know how I would do this within the controller?  Alternatively, is there another way to have the metaschemas available besides setting them on the controller?   Even though this is a draft route, I still need metaschema.  I used to return a RSVP.hash and return two models, but this is unreliable. 

As an aside, that API response you were getting above is exactly what I was getting on staging - for that and several other types of requests.  Not seeing this locally, however.

abought: One crude mechanism would be to only set the controller property after the promise was resolved, inside `.then`

addon/serializers/linked-node.js
pattisdr: I would think the data would need to be overridden - I mimicked Sam's changes in `addon/serializers/institution.js`.  The request for adding an institution to a node, or adding a linked-node to a collection look the same.

URL:          `/v2/collections/{collection_id}/relationships/linked_nodes/`
``` 
  {
                 "data": [{
                   "type": "linked_nodes",   # required
                   "id": <node_id>   # required
                 }]
               }
```

tests/dummy/app/routes/nodes/detail/draft-registrations/index.js
abought: Also: aren't metaschemas the possible kinds of registrations? (the question forms) 

It seems like what you really want is a separate or nested route: one page to show your existing registrations, and a separate but related one to create a new registration. That nested route could render as part of the page, but with its own model hook.

Not sure if it's a fit for your use case, but a useful option to have in your pocket.

abought:
I am seeing the following error on the console in functional testing. It's probably a blocker for this PR, but I'll continue other review while this is investigated.

`vendor.js:165665 Uncaught TypeError: r.createSnapshot is not a function`

And its closely related sibling, 
`vendor.js:165600 Uncaught TypeError: r.destroyRecord is not a function`

This happens on a project with many contributors. I removed two contributors, then added them again without refreshing the page. Usually this happened on the second contributor.

(or removed them, then added- the error message depends on the order of operations)

Sample test page:
http://localhost:4200/nodes/m9pv6

With two r+w contributors, `ramfn` and `z4esu`. (sorry for all the spam, dummy accounts!)

pattisdr:
I had been seeing this too when I added institution, and this commit fixed it https://github.com/CenterForOpenScience/ember-osf/pull/88/commits/4e7a2d87e9f0f4461df7fe601ff22a7b900b6e7e.

I hadn't tested with many contributors, I will look into that now.

abought:
> I hadn't tested with many contributors, I will look into that now.

Thanks. 

Followup question after this PR:

Is this a problem that we need to solve on a per model basis, or something that should be solved in the parent? As we write tests, we may want to consider how to DRY this code and support future additions.

addon/models/osf-model.js
```diff
@@ -20,20 +20,25 @@ export default DS.Model.extend(HasManyQuery.ModelMixin, {
         this._super(...arguments);
         this.set('_dirtyRelationships', Ember.Object.create({}));
     },
+    resolveRelationship(rel) {
```
abought: This seems to be a custom method of our own devising?

For the future, a docstring would be helpful for all newly added methods. We're accessing some pretty deep internals of ember data, and documenting the what and why would do a lot to help maintainability.

addon/adapters/osf-adapter.js
```diff
@@ -18,97 +20,254 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, DataAdapt
...
+     *
+     * @param {DS.Store} store
+     * @param {DS.Model} type
+     * @param {DS.Snapshot} snapshot
+     * @param {String} relationship
+     * @param {String} change
+     **/
+    _handleRelatedRequest(store, type, snapshot, relationship, change) {
+        var related = snapshot.record.get(`_dirtyRelationships.${relationship}.${change}`).map(r => r.createSnapshot());
+        // TODO(samchrisinger): will this have unintented side-effects for deletes/removes?
```
abought: Is this related to the errors we're seeing?

addon/serializers/linked-node.js
```diff
@@ -0,0 +1,16 @@
+import OsfSerializer from './osf-serializer';
+
+// TODO Dawn - a linked-node serializer is actually using the node model - Ember doesn't like this.
```
abought: Is this TODO still relevant?  Would we need a followup ticket to deal with any issues here?

I see we added a custom serializer.

addon/adapters/osf-adapter.js
samchrisinger: No, I think that TODO is now obsolete and can be removed.

addon/serializers/linked-node.js
pattisdr: I think so - unless you feel this method is suitable.  To add a project (linked-node) to a collection, the request is going to look a little different (mentioned above, needs to look like institutions).  So I created a serializer for the linked-node.  The only problem is, a linked-node is actually a node.  So the model type does not match the serializer.  So I could make a linked-node model, and just extend the node model, but I don't love that.

This technically works, but if you write a test for the linked-node serializer, it will complain because there's no linked-node model.  

Sam said just go with this for now and maybe look into it later.

abought:
Pass complete pending Dawn's investigation of issues with multiple requests in sequence.

This fixes several we saw previously- huzzah! Because it touches some undocumented private ember internals, my comments on the previous PR largely hold- the key test is that it works, and the best way to ensure will be putting a 1.0 version in front of users.

Main blockers are to fix the obvious breakages we know about, and remove any outdated TODOs (as these are a record of work needed).

We should follow up with documentation and tests in short order as a separate PR, while we remember what this code is for.

tests/dummy/app/routes/nodes/detail/draft-registrations/index.js
pattisdr: Yes - good point.  I am currently going to set the controller property after promise resolved per your first suggestion.  Drafts are going to need a large overhaul at some point and this can go in at that time (I don't think drafts are the big priority right now). My main reason for including this fix in this PR was so drafts as a relationship on the node could be tested.

abought: Agreed- drafts are in no way a blocker for this PR, and wouldn't be even if the API was behaving better. 

Just making the note while the code was fresh.

addon/models/osf-model.js
pattisdr: Yes - this code was used in the save method but it was pulled out to be a method on its own. This method was needed in the osf-adapter. The method returns a hasMany or BelongsToRelationship object that the addCanonicalRecord should be called on.  

I'll add the docstring!

pattisdr:
Okay - I think I've resolved the issues with multiple requests in sequence. The main problem has been the type of objects in the canonicalRecords - it was inconsistent. Thanks to @abought for pointing out a specific case I had missed.  Deleting in particular should be looked into in a follow-up PR.   

For future reference: right now I have some checks in the adapter to turn create snapshots from internalModels if not passed a snapshot.  But I think the root cause might be something we're not handling when we're deleting node links and deleting contributors (these call save on the node link and contributor and not the node, so maybe it's not hitting some of the custom code).

addon/models/osf-model.js
```diff
@@ -20,6 +20,12 @@ export default DS.Model.extend(HasManyQuery.ModelMixin, {
         this._super(...arguments);
         this.set('_dirtyRelationships', Ember.Object.create({}));
     },
+    /**
+     * Looks up relationship on model and returns hasManyRelationship
+     * or belongsToRelationship object.
+     *
+     * @param {String} rel, relationship on the model
+     **/
```
abought: Thanks!

(In the future we can also document return type etc. YUIDoc doesn't parse the source code- it just renders the things we choose to document explicitly)

abought:
I removed two contributors on a project, then tried to add them back without reloading. The result was the screenshot below.  Can you reproduce this?

It seems like a combination of dirty relationship tracking and how records are loaded into the store, maybe? 

URLs reference a staging project, using add contributors form in node detail page:
http://localhost:4200/nodes/m9pv6

- There seem to be some display anomalies with the new contributor record: the id displays, but not the person/name info, until the page reloads.
- For the second person I add back, I get an error about someone already being a contributor.

Here is the payload on the first request: (adding Sam)
```
https://staging-api.osf.io/v2/nodes/m9pv6/contributors/
Request Method:POST
Status Code:201 Created
Remote Address:104.130.12.42:443
```
```
{"data":[{"attributes":{"bibliographic":true,"permission":"read"},"type":"contributors","relationships":{"users":{"data":{"id":"x679r","type":"users"}}}}]}
```

After the second request (adding Dawn):
```
{"data":[{"attributes":{"bibliographic":true,"permission":"read"},"type":"contributors","relationships":{"users":{"data":{"id":"x679r","type":"users"}}}},{"attributes":{"bibliographic":true,"permission":"read"},"type":"contributors","relationships":{"users":{"data":{"id":"4hgq3","type":"users"}}}}]}
```

And a screenshot after the second request. 
screen shot 2016-06-30 at 7 17 42 pm

abought:
Ack- the branch didn't update when I thought it did. I double-checked commit hash before bed, and it worked with newest. So sorry for the inconvenience!

samchrisinger:
Thanks team for all the help with this one!

PR #89, opened 2016-06-29
Set up YUIDoc to work with ember-osf

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-81

# Purpose
Demonstrate ability to dynamically generate documentation for ember-osf modules.

screen shot 2016-06-29 at 9 44 38 am

# Notes for Reviewers
This is the same doc generator and convention as used by ember. Most of our docstrings already follow these conventions.

A separate PR would be required if we wanted to serve this via github pages. (with autodeployment code in this branch)

See sample deployment here: http://abought.github.io/ember-osf/

## How to use this
Install YUIDoc, then run it without arguments from the `ember-osf` root folder. Docs will be deposited in the `docs/` folder, as specified in yuidoc.json.

`npm -g install yuidocjs`
`yuidoc`

## TODO
Styling and improved text can move to a future ticket: get tooling out there and let the whole team make improvements.

- [x] Explore auto-deployment to Github pages

Here is ember's yuidoc.json file for comparison: https://github.com/emberjs/ember.js/blob/master/yuidoc.json

## Routes Added/Updated
None. Docs will appear in the new docs/ folder when command is run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
       "package.json",
       "bower_components",
       "dist",
+      "docs",
       "tmp",
       "vendor",
       "app/locales",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
   },
   "scripts": {
     "build": "ember build",
+    "docs": "yuidoc",
+    "docs-deploy": "yuidoc && ghp-import -r origin -p -b gh-pages docs/",
     "start": "ember server",
     "test": "ember test",
     "check-style": "./node_modules/jscs/bin/jscs ."
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 # compiled output
 /dist
+/docs
 /tmp
 
 # dependencies
--- a/addon/mixins/infinity-custom.js
+++ b/addon/mixins/infinity-custom.js
@@ -8,12 +8,12 @@ const assign = Ember.assign || Ember.merge;
  * For the most part, the API and semantics are identical to ember infinity, except that the means of configuring the store find method is more flexible
  *  (supporting relationship queries that do not operate via store methods)
 
- @class RouteMixin
+ @class InfinityCustomMixin
  @namespace EmberOSF
  @module ember-osf/mixins/infinity-custom
  @extends Ember.Mixin, InfinityRoute
  */
-const RouteMixin = Ember.Mixin.create(InfinityRoute, {
+export default Ember.Mixin.create(InfinityRoute, {
     /**
      * Repurpose an ember-infinity hook to override the method used for queries
      * @type {function}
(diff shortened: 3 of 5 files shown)


abought:
Moving forward for now with the python dependency. Below is a WIP bash script adapted from https://gist.github.com/domenic/ec8b0fc8ab45f39403dd ; it's failing on subsequent runs at the moment due to assumptions that do not apply outside travis. The python script is ugly but at least works so we can iterate process.

```
#!/usr/bin/env bash

#### Run documentation generation and push to github pages.
##   Pared down from https://gist.github.com/domenic/ec8b0fc8ab45f39403dd
## Supports argument "--remote" to provide name of the remote
set -e # Exit with nonzero exit code if anything fails

SOURCE_BRANCH="master"
TARGET_BRANCH="gh-pages"
REMOTE="origin"


while [ $# -gt 0 ]
do
    case "$1" in
        -r)  REMOTE="$2";;
    esac
    shift
done


function doCompile {
  yuidoc
}

# Save some useful information
REPO=`git config remote."$REMOTE".url`
SHA=`git rev-parse --verify HEAD`

# Clone the existing gh-pages for this repo into docs/
# Create a new empty branch if gh-pages doesn't exist yet (should only happen on first deploy)
git clone $REPO docs
cd docs
git checkout $TARGET_BRANCH || git checkout --orphan $TARGET_BRANCH
cd ..

# Clean out existing contents
rm -rf docs/**/* || exit 0

# Run our compile script
doCompile

# Now let's go have some fun with the cloned repo
cd docs

# If there are no changes to the compiled out (e.g. this is a README update) then just bail.
if [ -z `git diff --exit-code` ]; then
    echo "No changes to the output on this push; exiting."
    exit 0
fi

# Commit the "changes", i.e. the new version.
# The delta will show diffs between new and old versions.
git add .
git commit -m "Deploy to GitHub Pages: ${SHA}"


# Now that we're all set up, we can push.
git push $REMOTE $TARGET_BRANCH

```

package.json
```diff
@@ -8,6 +8,8 @@
   },
   "scripts": {
     "build": "ember build",
+    "docs": "yuidoc",
+    "docs-deploy": "yuidoc && ghp-import -r origin -p -b gh-pages docs/",
```
abought: Fun discovery: you can run `npm run <command_name> --arg1 value ...`, and actually pass arguments in after an npm command.

So there are ways to use a script, but make it configurable. :)

abought:
Looked into auto-generating more detailed documentation...

http://yui.github.io/yuidoc/syntax/
> IMPORTANT: YUIDoc only parses YUIDoc comment blocks, not source code. This keeps YUIDoc relatively simple and language agnostic. However, it also means you must declare everything to YUIDoc explicitly. A code snippet will not display as a "method" or "class" until you describe it as such. A corollary is that YUIDoc will never generate empty, "stub" doc entries for API members that lack comment blocks.

samchrisinger:
Awesome, thanks!

PR #90, opened 2016-06-30
[HOLD] Magic Guid Route

samchrisinger:
# Purpose

Initially R&D to explore using `Ember.Route.render` to render different subroutes under one wildcard route.

Updated to just add Guid and GuidReferent models plus making the User, Node, and File models subclass GuidReferent. This allows for a polymorphic `referent` relationship to exist on the Guid model, which in turn allows for doing: `store.find('guid', guid).get('referent')` to resolve a guid.
# Routes added

A `<guid>` route. Test this by grabbing a guid and visiting something like http://localhost:4200/xg8k5. 
screen shot 2016-07-19 at 3 25 29 pm

Right now invalid guids just 404 and break the whole page.

The Ember inspector confirms that the appropriate kind of model is loaded (no GuidReferents in the store).


tests/dummy/app/router.js
```diff
@@ -62,6 +62,9 @@ Router.map(function() {
     this.route('prereg');
     this.route('allnodes');
     this.route('usernodes');
+    this.route('guid', {
+        path: '/:guid'
```
abought: A small suggestion: this effectively functions as a catch-all route (picking up things like `/profile`), so as part of this we may want an error event that automatically tries the OSF with the same URL if a matching GUID segment is found.

That will be especially useful during incremental migration when some but not all resource types have been implemented in ember- if we link to a resource type it can''t handle, we'll let the OSF decide which page to serve.

tests/dummy/app/routes/guid.js
```diff
@@ -0,0 +1,26 @@
...
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, {
+    model(params) {
+        return this.store.findRecord('guid', params.guid).then(g => g.get('referent'));
+    },
+    renderTemplate(controller, model) {
+        let kind = model.constructor.modelName;
+        if (kind === 'node') {
+            this.render('nodes.detail.index', {
```
abought: This is interesting- this function seems to have some pretty powerful options under the hood. Being able to specify view and controller to be used during rendering sounds like a plus.

I'll be interested to see how this affects things like action bubbling and `thingFor` (eg what the files detail page thinks its parent model/params are). Certainly worth a look.

addon/adapters/guid.js
```diff
@@ -0,0 +1,17 @@
+import JSONAPIAdapter from 'ember-data/adapters/json-api';
+
+import config from 'ember-get-config';
+import DataAdapterMixin from 'ember-simple-auth/mixins/data-adapter-mixin';
+
+export default JSONAPIAdapter.extend(DataAdapterMixin, {
+    authorizer: 'authorizer:osf-token',
```
abought: Any reason for using JSONAPIAdapter instead of OsfAdapter here?

Two notes to bring this in line with other models:

Auth type should be configurable for cookie or token use:
`    authorizer: config['ember-simple-auth'].authorizer,

And instead of `DataAdapterMixin`, we may want to use `GenericDataAdapterMixin` to set XHR fields needed for cookie requests. (I'd love to make that unnecessary, but eh)

samchrisinger: Hmm. I wanted to avoid the OsfAdapter's extra magic, but in hindsight it shouldn't be a problem and additionally will help keep us away from divergent adapter implementations.

abought: *nods* If we want to simplify things, we could always strip down the base class and move the magic into `OsfAdapterDarkArtsMixin`. Composition might be preferable to duplication if it comes to that.

addon/models/guid-referent.js
```diff
@@ -0,0 +1,6 @@
+import OsfModel from './osf-model';
+
+/**
+ * Used only to support the polymorphic referent relationship on the Guid model
+ **/
+export default OsfModel.extend({});
```
abought: Spec question:

Looking at the OSF models, it several other kinds of information subclass `GuidStoredObject`. Will we need to add this base class to other models than Node/Registration, File, and User? Honest question; the definition vs usage in practice is a bit unclear.

Here is the class definition:

https://github.com/abought/osf.io/blob/723017aabd01af104ba316fa2390c575df88cd84/framework/guid/model.py#L53-L53
>     """Subclass of `StoredObject` that provisions a `Guid` for each new instance
    on save. When saving a `GuidStoredObject` for the first time, creates a new
    `Guid`, then assigns the primary key of the instance to the primary key of
    the `Guid`. Note: Subclasses should have a `StringField` primary key, since
    the key generated by the associated `Guid` will also be a string.
    """

And here is a list of things that directly inherit. Ones worth a look for ember-osf are noted in bold:
- User
- ~~Badge~~
- *NodeWikiPage*
- *Comment*
- Node

Additionally. `StoredFileNode` calls `Guid.generate` optionally in certain situations.

abought:
Pass complete. Basic test route works as expected, and confirmed that records are entering the store in the correct place. 

It's possible that wikis need to inherit from `guid-referent` as well. I don't recall ever seeing a wiki guid URL, but it seems to be a thing supported by the code and therefore possibly for URLs in the wild. 

Comments have GUIDs, but don't seem to implement their own `deep_url` method allowing them to be resolved in the OSF via guid route. Supporting them for GUID resolution does not therefore seem to be necessary.

addon/models/guid-referent.js
samchrisinger: Hmm. This actually points out some holes in the v2 implementation... looking into this some more.

abought:
Alternate methods were put in place as part of preprints, so we can close this ticket to clear the backlog a bit.

brianjgeiger:
It should be noted that the method put in for Preprints only works if you have only one kind or object. So a single ember app could render a preprint or a node or a registration but not more than one of those things in the app.

PR #91, opened 2016-07-01
Feature/Some Mixin Docstrings [EOSF-89]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-89

# Purpose

Adds docstrings to the registration-actions/node actions mixins and taggable-mixin.  Adds docstrings to dummy app controllers/route actions for drafts, controllers, profile.

# Notes for Reviewers

### DIFF
--- a/addon/mixins/node-actions.js
+++ b/addon/mixins/node-actions.js
@@ -4,19 +4,41 @@ export default Ember.Mixin.create({
     node: null,
     model: null,
     _node: Ember.computed.or('node', 'model'),
-
-    /** Maps all node contributors to format {contribID: Contributor}
-     **/
+    /**
+    * Helper method that maps all node contributors to format {contribId: Contributor}
+    *
+    * @method _generateContributorMap
+    * @param {Object} contributors All contributor relationships on the node
+    * @return {Object} Returns a contributor map of the id to the contributor record
+    */
     _generateContributorMap(contributors) {
         var contribMap = {};
         contributors.forEach(contrib => contribMap[contrib.id] = contrib);
         return contribMap;
     },
+    /**
+    * Helper method that affiliates an institution with a node.
+    *
+    * @method _affiliateNode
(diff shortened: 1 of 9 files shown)


addon/mixins/node-actions.js
```diff
@@ -4,19 +4,41 @@ export default Ember.Mixin.create({
...
     model: null,
     _node: Ember.computed.or('node', 'model'),
-
-    /** Maps all node contributors to format {contribID: Contributor}
-     **/
+    /**
+    * Helper method that maps all node contributors to format {contribId: Contributor}
+    *
+    * @method _generateContributorMap
+    * @param {Object} contributors All contributor relationships on the node
```
abought: Should this type be `Array` instead of object?

pattisdr: You're right, I guess contributors are an array.  Every place where it was a resource instead of a string/boolean/date, I just called it an object.  I will go through and see where else I made this mistake.

addon/mixins/registration-actions.js
```diff
@@ -22,6 +29,15 @@ export default Ember.Mixin.create({
         }
         return currentMetadata;
     },
+    /**
+    * Build the request payload used when creating a registration from a draft.
+    *
+    * @method _generateRegistrationPayload
+    * @param {String} draft ID of the draft registration
+    * @param {String} registrationChoice Either "immediate" or "embargo".
+    * @param {Date} liftEmbargo, if registrationChoice === "embargo". liftEmbargo should be the date to lift the embargo.
```
abought: Unbelievably trivial: there are a few trailing commas that are probably not needed throughout. (and one period that could be a comma)

abought:
Two small notes, but overall this PR is a welcome thing.

pattisdr:
Thanks for the quick review @abought, changes made.

abought:
Looks good!

PR #97, opened 2016-07-08
Add temporary fix for Comment.replies serialization

samchrisinger:
# Purpose 

See: https://openscience.atlassian.net/browse/OSF-6646

Comment serialization for the replies relationship is incorrect, which confuses ember-data (which uses data/relationships/<name>/links/related to determine if a related field has data) This PR temporarily patches the serialization client side until we can fix the API.

### DIFF
--- a/addon/serializers/comment.js
+++ b/addon/serializers/comment.js
@@ -1,7 +1,8 @@
+import Ember from 'ember';
 import OsfSerializer from './osf-serializer';
 
 export default OsfSerializer.extend({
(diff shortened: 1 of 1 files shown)


abought:
Thanks. Can we add a link to this PR to the tracking issue for any APIv2 changes? That way we'll know exactly what code we can remove, when the time comes.

samchrisinger:
@abought I have no plans right now to do any OSF development for this. But we have https://openscience.atlassian.net/browse/EOSF-95 to track the future removal of this fix.

PR #98, opened 2016-07-08
Display styled components in consuming apps

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-94

# Purpose
In the case where ember-osf components require styling, these styles should be available to consuming apps.

This PR has two incremental goals:
1. Make the styles available at all, in a drop-in manner (as dependencies, not devDependencies, where possible)
2. Make dependencies configurable: do not expose dependencies that would increase the size of the built assets unnecessarily.

screen shot 2016-07-12 at 2 13 23 pm

# Notes for Reviewers
## Future directions
At some point, we may wish to consider splitting models and styled/functional components into separate ember-addons, in order to keep the list of dependencies minimal. Not every use case will require our pre-built interactive widgets.

## Major asset categories to consider
- Libraries (bootstrap, font-awesome)
- Component styles
- Third party libraries required as a dependency 

## TODO

- [x] Gravatar is too big in dummy app and obscures utility navbar
screen shot 2016-07-12 at 3 48 40 pm

- [x] Make shared style libraries available (bootstrap, FA)
- Verify styling of components:
  - [x] osf-navbar
  - [x] osf-footer
  - [x] osf-mode footer 
  - [x] file-browser
  - [x] file-widget
  - [x] search-dropdown
  - [x] sign-up

## Routes Added/Updated
None. However, let me know if dummy app stops working or shows inappropriate styling.


abought:
Documentation of what you need to work with styles will follow, but I'd like to get the companion OSF PR together first so we have a common point to work from.

samchrisinger:
@abought is this rtm? Are we still not getting file-browser and file-widget styles?

abought:
Verifying those last two in the consuming app now, and then will move this ticket forward.

The gravatar icon is a bit too large (see screenshot) but I'm tempted to push that one off.

PR #101, opened 2016-07-14
Add unit tests for OsfAdapter

samchrisinger:
## Ticket
https://openscience.atlassian.net/browse/EOSF-78

# Purpose

The OsfAdapter does quite a bit of heavy lifting that bypasses some Ember internals. We need unit tests to document and prevent regressions in this code.

# Notes for Reviewers

Should mostly be added tests. Some bugs might get fixed in the process.

### DIFF
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -162,13 +162,9 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
      * @param {Boolean} isBulk
      **/
     _deleteRelated(store, snapshot, deletedSnapshots, relationship, url, isBulk = false) { // jshint ignore:line
-        if (isBulk) {
-            return this._removeRelated(...arguments).then(() => {
-                deletedSnapshots.forEach(s => s.record.unloadRecord());
-            });
-        } else {
-            return Ember.RSVP.allSettled(deletedSnapshots.map(r => r.record.unloadRecord()));
-        }
+        return this._removeRelated(...arguments).then(() => {
+            deletedSnapshots.forEach(s => s.record.unloadRecord());
+        });
     },
     /**
      * A helper for making _*Related requests
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -103,7 +103,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
     _addRelated(store, snapshot, addedSnapshots, relationship, url, isBulk = false) {
         return this._doRelatedRequest(store, snapshot, addedSnapshots, relationship, url, 'POST', isBulk).then(res => {
             addedSnapshots.forEach(function(s) {
-                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(s);
+                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(s.record);
             });
             return res;
         });
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -125,7 +125,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
             var relatedType = singularize(snapshot.record[relationship].meta().type);
             res.data.forEach(item => {
                 var record = store.push(store.normalize(relatedType, item));
-                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(record._internalModel.createSnapshot());
+                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(record);
             });
             return res;
         });
@@ -145,7 +145,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
      **/
     _removeRelated(store, snapshot, removedSnapshots, relationship, url, isBulk = false) {
         return this._doRelatedRequest(store, snapshot, removedSnapshots, relationship, url, 'DELETE', isBulk).then(res => {
-            removedSnapshots.forEach(s => snapshot.record.resolveRelationship(relationship).removeCanonicalRecord(s));
+            removedSnapshots.forEach(s => snapshot.record.resolveRelationship(relationship).removeCanonicalRecord(s.record));
             return res || [];
         });
     },
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -238,7 +238,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
     _handleRelatedRequest(store, type, snapshot, relationship, change) {
         var related = snapshot.record.get(`_dirtyRelationships.${relationship}.${change}`).map(function(r) {
             if (r._internalModel) {
-                return r;
+                return r._internalModel.createSnapshot();
             }
             return r.createSnapshot();
         });
--- a/addon/adapters/osf-adapter.js
+++ b/addon/adapters/osf-adapter.js
@@ -83,7 +83,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
                 url: url
             }
         }).then(res => {
-            createdSnapshots.forEach(s => snapshot.record.resolveRelationship(relationship).addCanonicalRecord(s));
+            snapshot.record.resolveRelationship(relationship).addCanonicalRecord(s.record);
             return res;
         }));
     },
@@ -261,12 +261,12 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
 
         var response;
         response = adapter[`_${change}Related`](
-                    store,
-                    snapshot,
-                    related,
-                    relationship,
-                    url,
-                    allowBulk
+            store,
+            snapshot,
+            related,
+            relationship,
+            url,
+            allowBulk
         );
         return response;
     },
--- a/addon/mixins/node-actions.js
+++ b/addon/mixins/node-actions.js
@@ -121,7 +121,7 @@ export default Ember.Mixin.create({
         addContributor(userId, permission, isBibliographic) {
             var node = this.get('_node');
             var contributor = this.store.createRecord('contributor', {
-                id: userId,
+                id: `${node.get('id')}-${userId}`,
                 permission: permission,
                 bibliographic: isBibliographic
             });
--- a/tests/unit/adapters/osf-adapter-test.js
+++ b/tests/unit/adapters/osf-adapter-test.js
@@ -8,6 +8,7 @@ import FactoryGuy, {
 } from 'ember-data-factory-guy';
 
 import DS from 'ember-data';
+import JSONAPIAdapter from 'ember-data/adapters/json-api';
 import OsfAdapter from 'ember-osf/adapters/osf-adapter';
 
 moduleFor('adapter:osf-adapter', 'Unit | Adapter | osf adapter', {
@@ -24,7 +25,7 @@ moduleFor('adapter:osf-adapter', 'Unit | Adapter | osf adapter', {
 });
 
 test('#buildURL appends a trailing slash if missing', function(assert) {
-    var url = 'https://api.osf.io/v2/users/me';
+    var url = 'http://localhost:8000/v2/users/me';
     this.stub(
         DS.JSONAPIAdapter.prototype,
         'buildURL',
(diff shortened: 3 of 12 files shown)


addon/adapters/osf-adapter.js
```diff
@@ -125,7 +125,7 @@ export default DS.JSONAPIAdapter.extend(HasManyQuery.RESTAdapterMixin, GenericDa
             var relatedType = singularize(snapshot.record[relationship].meta().type);
             res.data.forEach(item => {
                 var record = store.push(store.normalize(relatedType, item));
-                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(record._internalModel.createSnapshot());
+                snapshot.record.resolveRelationship(relationship).addCanonicalRecord(record);
```
abought: I like that we're not relying on a chain of multiple private methods- was this found to be simply unnecessary? (I don't see a larger refactoring that changes the type of data being passed in)

addon/mixins/node-actions.js
```diff
@@ -121,7 +121,7 @@ export default Ember.Mixin.create({
         addContributor(userId, permission, isBibliographic) {
             var node = this.get('_node');
             var contributor = this.store.createRecord('contributor', {
-                id: userId,
+                id: `${node.get('id')}-${userId}`,
```
abought: Good catch!

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+        'serializer:node',
+        'service:session',
+        'transform:links', 'transform:embed'
+    ],
+    beforeEach() {
+        manualSetup(this.container);
+    }
+});
+
+test('#buildURL appends a trailing slash if missing', function(assert) {
```
abought: Small suggestion based on past regressions: can you please also add a test for

"Does not append a trailing slash if one is already present"

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+            'user',
+            'me',
+            user._internalModel.createSnapshot(),
+            `${verb}Record`
+        );
+        assert.equal(url, result);
+    });
+});
+
+test('#buildURL uses snapshot.adapterOptions.url if available', function(assert) {
```
abought: Good!

I notice that there is another `adapterOptions` key called `nested`, which does not yet have a test, and some of our custom code checks for the flag: https://github.com/abought/ember-osf/blob/b52986fadfdbdc1e2e5d5e0a79e9621ad5b9f61d/addon/adapters/osf-adapter.js#L82-L82

Should we have a test for that flag as well? (can't seem to find good documentation on ember's built-in keys)

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+                url: url
+            }
+        }),
+        'createRecord'
+    );
+    assert.equal(url, result);
+});
+
+test('#_buildRelationshipURL uses relationshipLinks', function(assert) {
+    let url = 'https://api.osf.io/v2/users/me/nodes/';
```
abought: This seems fairly similar to the kind of relationship URL that ember would create by default. 

To ensure the test isn't passing by coincidence, should we perhaps add a second assertion (eg that user ID != 'me')?

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+    Ember.run.begin();
+    let contributors = [
+        store.createRecord('contributor', {
+            title: 'Foo'
+        }),
+        store.createRecord('contributor', {
+            title: 'Bar'
+        })
+    ];
+    Ember.run.end();
```
abought: What is the advantage of manually specifying begin and end? Alternately, you can run an entire block as `Ember.run(function() {})`

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+
+    var doRelatedStub = this.stub(OsfAdapter.prototype, '_doRelatedRequest', () => {
+        return new Ember.RSVP.Promise(resolve => resolve());
+    });
+    var relation = node.resolveRelationship('affiliatedInstitutions');
+    relation.hasLoaded = true;
+    var addCanonicalStub = this.stub(relation, 'addCanonicalRecord');
+
+    Ember.run(() => {
+        node.save().then(() => {
```
abought: Normally `node.save` does ajax work. Where *does* that request go during a unit test scenario?

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+    var pushStub = this.stub(store, 'push', () => contrib);
+    var normalizeStub = this.stub(store, 'normalize');
+
+    Ember.run(() => {
+        node.save().then(() => {
+            assert.ok(doRelatedStub.calledOnce);
+            assert.ok(addCanonicalStub.calledOnce);
+            assert.ok(pushStub.calledOnce);
+            assert.ok(normalizeStub.calledOnce);
+        }, () => {
```
abought: Instead of passing two arguments to `.then`, the second failure handler can be specified as `.then().catch()`. My impression is that using separate handlers is preferred over using two arguments to `.then`, partly because it helps ensure that promises are chained correctly- eg to ensure that `.catch` handles errors in the first request *or* any that happen during the success callback.

(not 100% sure about this, as some articles are a bit old)

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+
+    var addCanonicalStub = this.stub();
+    this.stub(node, 'resolveRelationship', () => {
+        return {
+            addCanonicalRecord: addCanonicalStub
+        };
+    });
+
+    Ember.run(() => {
+        node.save().then(() => {
```
abought: See note above about `.then / .catch` semantics.

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+    var rel = node.resolveRelationship('affiliatedInstitutions');
+    var removeCanonicalStub = this.stub(rel, 'removeCanonicalRecord', removeCanonicalStub);
+    rel.hasLoaded = true;
+
+    Ember.run(() => {
+        node.save().then(() => {
+            assert.ok(doRelatedStub.calledOnce);
+            assert.ok(removeCanonicalStub.calledOnce);
+            assert.ok(removeCanonicalStub.calledWith(inst));
+        }, () => {
```
abought: See previous note on `.then/.catch`

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+        child._internalModel.createSnapshot(),
+        'children',
+        'https://api.osf.io/v2/nodes/foobar/children/',
+        'POST'
+    );
+    var data = mockAjax.args[0][2].data.data;
+    assert.equal(data.attributes.title, child.get('title'));
+});
+
+test('#_handleRelatedRequest makes correct calls for each change argument', function(assert) {
```
abought: Do we have a test to ensure that a record is correctly saved when we change both relationships *and* the original record? Eg if we add contributors and also change node title, we want to ensure that all changes get saved.

This is about the closest I could find (it checks that multiple operations are performed), but it doesn't seem to check anything about the original record.

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -1,12 +1,401 @@
...
+        );
+        assert.ok(relatedStub.called);
+        assert.deepEqual(
+            relatedStub.args[0][2].map(s => s.id),
+            changes[verb].map(r => r.get('id'))
+        );
+    }
+});
+
+test('#_handleRelatedRequest checks if relationship supports bulk', function(assert) {
```
abought: Should this also check that the correct ajaxOptions (`ret.contentType`) are being set?

It's possible that deep bulk support is slated for a separate PR, in which case exhaustive bulk testing could wait.

abought:
Pass complete. The existing tests look ok. 

I have a few small suggestions for cleanup of promise semantics before this goes ahead, as well as some judgment call questions about potential things that need to be tested. No major blockers though, and these changes work in light functional testing.

tests/unit/adapters/osf-adapter-test.js
samchrisinger: The stubbing of `_doRelatedRequest` prevents and ajax calls from being made.

tests/unit/adapters/osf-adapter-test.js
samchrisinger: The [RSVP source code comments](https://github.com/tildeio/rsvp.js/blob/master/lib/rsvp/promise.js#L107) use the pattern above, and I've experienced weirdness with using the `catch` callback (maybe a transpilation issue?). Additionally, the usage here is pretty constrained, i.e. I'm positive no other code is depending on the resolution of this promise.

So unless you're really sure and feel strongly about this, I'd prefer to leave it as is.

addon/adapters/osf-adapter.js
samchrisinger: I think the old version only worked by luck. Looking [at the source](https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/relationship.js#L66) it seems a plain record and not a snapshot is needed here.

tests/unit/adapters/osf-adapter-test.js
abought: Hmm. I've definitely noticed weaker ES6 support in tests before (eg phantomJS is headless, and often outdated, webkit- at times I will see tests pass in chrome but fail in phantom when I use `ember test --server`).

I'd strongly encourage separate methods in production code, but no sense fighting tooling for unit tests. Ok to leave as is, then.

abought: Useful article on promises for future reference:
http://stackoverflow.com/questions/24662289/when-is-thensuccess-fail-considered-an-antipattern-for-promises


The bluebird documentation refers to `.then(success, fail)` as an [antipattern](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-thensuccess-fail-anti-pattern), for subtle reasons.

tests/unit/adapters/osf-adapter-test.js
samchrisinger: The flag is used here: https://github.com/CenterForOpenScience/ember-osf/blob/ae56328ff79bf8cb71041b21bf095d110282a36a/addon/models/osf-model.js#L51. So the test you're looking for is 
> records saved with #_createRelated include the `nested` adaperOption

?

abought: a) That the option is included, and,
b) That it does whatever it's supposed to do

From an initial read, I wasn't 100% clear on the functionality this is intended to provide, and it would be good to have an authoritative source on what and how this should work.

tests/unit/adapters/osf-adapter-test.js
samchrisinger: Added: https://github.com/CenterForOpenScience/ember-osf/pull/101/commits/929bdd4afa966375ad13910000710aa4147e65c9#diff-34b9b2293f5a3ec1d6c56ad1b433a873R48

tests/unit/adapters/osf-adapter-test.js
samchrisinger: Added: https://github.com/samchrisinger/ember-osf/blob/ef32d259c5c911191b5579c41df93e0806368a1a/tests/unit/adapters/osf-adapter-test.js#L198

We need to test the other side of the interface (in OsfModel) in a followup

tests/unit/adapters/osf-adapter-test.js
samchrisinger: Changed: https://github.com/samchrisinger/ember-osf/blob/ef32d259c5c911191b5579c41df93e0806368a1a/tests/unit/adapters/osf-adapter-test.js#L128

tests/unit/adapters/osf-adapter-test.js
samchrisinger: Added: https://github.com/samchrisinger/ember-osf/blob/ef32d259c5c911191b5579c41df93e0806368a1a/tests/unit/adapters/osf-adapter-test.js#L486

tests/unit/adapters/osf-adapter-test.js
samchrisinger: Added: https://github.com/samchrisinger/ember-osf/blob/ef32d259c5c911191b5579c41df93e0806368a1a/tests/unit/adapters/osf-adapter-test.js#L519

samchrisinger:
@abought changes made

tests/unit/adapters/osf-adapter-test.js
```diff
@@ -399,3 +482,44 @@ test('#_handleRelatedRequest checks if relationship supports bulk', function(ass
...
 });
+
+test('#updateRecord handles both dirtyRelationships and the parent record', function(assert) {
+    let adapter = this.subject();
+
+    this.inject.service('store');
+    let store = this.store;
+
+    let node = FactoryGuy.make('node');
+    Ember.run(() => node.set('title', 'The meaning of life'));
```
abought: This test does seem to check that the dirty relationships are updated. Can you point out where it verifies that the new title is also persisted correctly?

samchrisinger: I'm not especially keen on attempting to check that a model gets 'saved' in a unit test. Stubbing the super call (L502) adequate here.

PR #105, opened 2016-07-21
[hold] Convert `tags-widget` to ember

el1t:
## Ticket

N/A
# Purpose

`tags-widget` was previously implemented through a jquery plugin, which did not take advantage of Ember's built-in data binding and was far less extensible. This new implementation also has a few features:
- Populates input with deleted tag text for quick editing
- Supports validation via `pattern` regex
- (True) Auto-expanding `input`

This also removes two dependencies, `jquery.tagsInput` and `typeahead.js`, the latter of which was unused.

screen shot 2016-07-21 at 10 44 19 am
# Notes for Reviewers
- `/^.{1,128}$/` seems to be the default validation from osf.io
- Styling repeats some from bootstrap because it is not explicitly imported
- Does not support typeahead, but I do not think it was implemented in the previous version
- Didn't implement `canEdit` functionality, if needed I can do that
## Routes Added/Updated

None

### DIFF
--- a/addon/components/tags-widget/component.js
+++ b/addon/components/tags-widget/component.js
@@ -9,62 +9,83 @@ import layout from './template';
 /**
  * Allow the user to view and manage tags.
  * See TaggableMixin for controller actions that can be used with this component.
- *
- * For more information on configuration options, see documentation for [jquery-tags-input](https://github.com/xoxco/jQuery-Tags-Input).
+ * Pattern is a regular expression to validate input against.
  *
  * ```handlebars
  * {{tags-widget
- *   addATag=(action 'addATag')
- *   removeATag=(action 'removeATag')
- *   tags=model.tags}}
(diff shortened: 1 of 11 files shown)


samchrisinger:
@el1t tests are failing

samchrisinger:
@el1t on a Node with no tags, I get:

screen shot 2016-08-03 at 11 24 57 am

samchrisinger:
My guess is L35-37:
```javascript
    if (!_ember['default'].isArray(tags)) {
        this.set('tags', tags ? [tags] : []);
    }
```

needs to be:
`     this.set('tags', tags ? [tags] : Ember.A());`

el1t:
I am concerned about typeahead though—was it being used before? It may be more difficult to add typeahead through this approach.

laurenbarker:
Closing since this is now stale.

PR #108, opened 2016-07-27
[WIP] Fix regressed items - update contributors, multiple sequential requests

pattisdr:
This branch needed for node contributor page demo to work!

Needed for https://openscience.atlassian.net/browse/EOSF-118

# Purpose

Quick fix, should be improved later.  Ember-OSF addon/serializers/contributor is used for both adding and updating a contributor.   For adding a contributor, the user id needs to be in the request under relationships. For updating a contributor, we need to use the new contributor id format (node-id:user-id).

Picks appropriate id format and sends in with contributor request.

# Notes for Reviewers
The API accepts an incorrectly formatted request for updating contributor, which is why the same Ember serializer can be used for both adding/updating.  So might need to tighten constraints for relationship requests API side. Also, more work needs to be done on the serializers ember-side in general to make sure requests are formatted correctly (even though DRF is ignoring extra fields).

## Routes Added/Updated

None


abought:
Nice to see some adapter improvements!

The adapter represents some complex code with many subtle requirements. Can we add some tests?
1. Unit tests for the new functionality, and,
2. Regression tests to ensure that the bug we saw doesn't happen again

abought:
I am still seeing a related issue where adding 3 contributors from the contributors modal (OSF-ember project) causes 9 requests to be sent. 

That may be outside scope of this PR / available labor time, though certainly a useful fix for supporting the contribs page.

PR #110, opened 2016-08-03
Fix error where cookie auth beforeModel hook failed

abought:
## Ticket
https://openscience.atlassian.net/browse/PREP-87

# Purpose
Fix issue where (using cookie auth) the index page failed to load because the user was logged out.

When user was not logged in, beforeModel hook returned a rejected promise and failed. Catch that error and do nothing, so hook can pass through. (this just gets credentials if possible; it doesn't impose a hard requirement that the user be logged in)

BeforeModel hook (and ember simple auth) expect an ember promise, but receives a jquery object with slightly different semantics. Massage the promise type so things work correctly.

# Notes for Reviewers

## Routes Added/Updated
Anything using cookie auth: application route, eg in osf-ember and preprints app.

### DIFF
--- a/addon/authenticators/osf-cookie.js
+++ b/addon/authenticators/osf-cookie.js
@@ -1,3 +1,5 @@
+import Ember from 'ember';
+
 import Base from 'ember-simple-auth/authenticators/base';
 import config from 'ember-get-config';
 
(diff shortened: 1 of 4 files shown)


pattisdr:
Reviewing this morning -

pattisdr:
@abought - Here's what I am running:
- ember-osf: this branch, fix/prep87-auth-errors
- ember-preprints: feature/prep87-app-integration
- osf feature branch ember-preprints

I am trying to make sure I am testing the right thing:
 - I am logged out.  If I navigate to localhost:5000/preprints, the page loads with a login button in the navbar, as expected.  Is this due to the changes in this PR?

pattisdr:
Confirmed behavior works as expected; looks good to me.

PR #111, opened 2016-08-09
Feature/OSF Paginator and unregisteredContributor Action

pattisdr:
## Ticket

Ember-osf half of https://openscience.atlassian.net/browse/PREP-8.  
❗ Needed for https://github.com/CenterForOpenScience/ember-preprints/pull/19

# Purpose

Ember-osf items needed for author management of preprints.  Majority of PR has been removed and made a PR to ember-preprints.

Items included here - 
- osf-pagination component, behaves like the pagination component we use for search results for contributors in the OSF
- adds addUnregisteredContributor component to NodeActions mixin
- addContributor method returns contributor instead of node promise.
- action for reordering contribs

# Notes for Reviewers


addon/components/osf-paginator/component.js
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            paginator.pushObject('...');
+            for (var l = currentPage - 1; l <= currentPage + 1; l++) {
+                paginator.pushObject(l + 1);
+            }
+            paginator.pushObject('...');
+        }
+
+        paginator.push(pages);
+        paginator.push('>');
```
abought: Am I correct in assuming that most of this code is borrowed from the current OSF paginator implementation?

We'll probably want to rework this in the future, though I do recognize that for now we just need something to move forward with.

addon/components/osf-paginator/component.js
```diff
@@ -0,0 +1,68 @@
...
+        return paginator;
+    }),
+    actions: {
+        findResults(query, page) {
+            this.sendAction('fetchResults', query, page);
```
abought: I noticed that you were looking at closure actions, and I think they would be good to use here. The big advantage: you don't want to change the page number if `fetchResults` fails with an error.

(sendAction just fires the action as is with no way of knowing whether it worked. Closure actions let you check a return value, if called using `this.attrs.closureActionName(...arguments)`.

addon/components/osf-paginator/component.js
pattisdr: Yes, my goal here was to turn the OSF paginator into an ember component.

addon/mixins/node-actions.js
```diff
@@ -129,6 +129,26 @@ export default Ember.Mixin.create({
...
+         */
+        addUnregisteredContributor(fullName, email, permission, isBibliographic) {
+            var _this = this;
+            var user = this.store.createRecord('user', {
+                fullName: fullName,
+                username: email
+            });
+            // After user has been saved, add user as a contributor
+            return user.save().then(function(user) {
+                _this.send('addContributor', user.id, permission, isBibliographic);
```
abought: You shouldn't need `_this = this`; just use an arrow function in your promise handler instead.

`.then(user => result)`

(Arrow functions don't redefine the word "this")

addon/components/osf-paginator/component.js
```diff
@@ -0,0 +1,68 @@
+import Ember from 'ember';
+import layout from './template';
+
+// Ember component that is very similar to osf/website/static/js/paginator.js
```
abought: Even though I don't want us to get too attached to the old OSF pagination code, we should document the intended API (including parameters required to use it), to support future refactoring.

Otherwise, there are variables used (but not defined) in this code- it should be clear where data comes from.

Most of our other components have docstrings that can be used as examples. Can you please add something similar here following the pattern set out elsewhere in the ember-osf codebase?

pattisdr: Yes thanks for the reminder.

abought:
I'm deliberately not going too in depth on review in the interests of deadlines, but at minimum, can you please add documentation, and change this to use closure actions?

The code doesn't need to look pretty during the demo, but proper error handling will help avoid user-facing glitches, like if the buttons get stuck and don't do things.

pattisdr:
Addition: Add contributor method returns contributor instead of node promise.

pattisdr:
Addition: Add action for reordering contributors.

addon/mixins/node-actions.js
```diff
@@ -144,7 +142,14 @@ export default Ember.Mixin.create({
             });
             // After user has been saved, add user as a contributor
             return user.save().then(user => {
-                this.send('addContributor', user.id, permission, isBibliographic);
+                var node = this.get('_node');
```
abought: I'm a little concerned about the duplicate code here, but we can dig into that later once other higher priorities have been cleared up.

pattisdr: Also, this whole method might change this week if APIv2 changes for adding unregistered contributors go through, allowing contributors to be created with an email/fullname combo instead of the userid. If that is the case, I will put in a follow-up PR.

PR #112, opened 2016-08-12
[WIP] Feature/Return Contributor Save instead of Node Save [EOSF-141]

pattisdr:
## Ticket

❗ Not ready exploratory only
https://openscience.atlassian.net/browse/EOSF-141

# Purpose

Calling .save() on a node is masking errors when changing relationships.  Noticed in preprints author management PR https://github.com/CenterForOpenScience/ember-preprints/pull/19

*Example: Updating contributors.*
We have an updateContributors method on the NodeActionsMixin that bulk updates contributors.  We update a contributor by setting the contributor record with the new permission/bibliographic information contributor.set('permission', 'read') and then call node.save().  If I stop my API server (so the request will fail), the promise returned from node.save() is fulfilled.  There's no indication that the request didn't go through except in the console.

However, if I call contributor.save() with my API server stopped, the promise rejects as I would expect it to.  

# Notes for Reviewers

❗ Just early steps...if we actually move to returning contributor.save() instead of node.save() when modifying contributors, I imagine we will need some modifications to dirty relationships.

### DIFF
--- a/addon/mixins/node-actions.js
+++ b/addon/mixins/node-actions.js
@@ -113,8 +113,8 @@ export default Ember.Mixin.create({
          *
          * @method addContributor
          * @param {String} userId ID of user that will be a contributor on the node
-         * @param {String} permission User permission level. One of "read", "write", or "admin". Default: "write".
-         * @param {Boolean} isBibliographic Whether user will be included in citations for the node. "default: true"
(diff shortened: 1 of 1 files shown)


pattisdr:
Closing in favor of @chrisseto's work.

PR #115, opened 2016-08-17
Manually handle relations

chrisseto:


addon/models/node.js
```diff
@@ -139,5 +139,87 @@ export default OsfModel.extend(FileItemMixin, {
...
+        }).then(resp => {
+            child.unloadRecord();
+            this.store.pushPayload(resp);
+            let created = this.store.peekRecord('node', resp.data.id);
+            this.get('children').pushObject(created);
+            return created;
+        });
+    },
+
+    addUnregisteredContributor(fullName, email, permission, isBibliographic, index=9999) {
```
samchrisinger: `index=Number.MAX_SAFE_INTEGER`?

chrisseto: 9007199254740991/10 would change

addon/models/node.js
```diff
@@ -139,5 +139,87 @@ export default OsfModel.extend(FileItemMixin, {
...
+        let user = this.store.createRecord('user', {
+            fullName: fullName,
+            username: email
+        });
+        // After user has been saved, add user as a contributor
+        return user.save().then(user => this.addContributor(user.id, permission, isBibliographic, index));
+
+    },
+
+    addContributor(userId, permission, isBibliographic, index=Number.MAX_SAFE_INTEGER) {
```
pattisdr: When I call addContributor there are two items that get pushed into the store, one with the user id and one with the correct contributor id.
screen shot 2016-08-17 at 9 58 22 am

addon/models/node.js
```diff
@@ -139,5 +139,87 @@ export default OsfModel.extend(FileItemMixin, {
...
+            this.store.pushPayload(resp);
+            let created = this.store.peekRecord('contributor', resp.data.id);
+            this.get('contributors').pushObject(created);
+            return created;
+        });
+    },
+
+    updateContributor(contributor, permissions, bibliographic) {
+        if (permissions === '')
+            contributor.set('permission', permissions);
```
pattisdr: Will not update contributor.  Checking to see if permission or bibliographic is blank string instead of not a blank string.

addon/mixins/node-actions.js
```diff
@@ -178,15 +144,21 @@ export default Ember.Mixin.create({
...
-            }
-            for (let contributorId in bibliographicChanges) {
-                contributorMap[contributorId].set('bibliographic', bibliographicChanges[contributorId]);
-            }
-            return node.save();
+            return this.get('_node').updateContributors(...arguments);
+        },
+
+        /**
+         * Update contributors of a node. Makes a bulk request to the APIv2.
```
pattisdr: Docstring needs updating.  Not a bulk request.

pattisdr:
I know your PR didn't touch this, but removing a contributor is also not taking the contributor out of the store
screen shot 2016-08-17 at 10 13 08 am
. I do not know if this is important.

addon/models/node.js
```diff
@@ -139,5 +139,87 @@ export default OsfModel.extend(FileItemMixin, {
...
+        return user.save().then(user => this.addContributor(user.id, permission, isBibliographic, index));
+
+    },
+
+    addContributor(userId, permission, isBibliographic, index=Number.MAX_SAFE_INTEGER) {
+        let contrib = this.store.createRecord('contributor', {
+            // Original code used the line below.
+            // Serialize does something weird I guess
+            // id: `${this.get('id')}-${userId}`,
+            id: userId,
```
pattisdr: Contributor ids are now a combination of node-id and user id.  nodeid-userid.

pattisdr: Before when contributors were being created, it ran through osf-adapter/create-related, and we added 'create' to adapterOptions here https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/adapters/osf-adapter.js#L81-L84.

Then when it got to the contributor serializer, we would see the 'create' attached in previous step and know to use userId instead of contributor id. https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/serializers/contributor.js#L9-L10

The same contributor serializer is handling both create and update requests, but create method needs user id and update method needs contributor id.

addon/models/node.js
```diff
@@ -139,5 +139,87 @@ export default OsfModel.extend(FileItemMixin, {
...
+            bibliographic: isBibliographic,
+        });
+
+        return this.store.adapterFor('contributor').ajax(this.get('links.relationships.contributors.links.related.href'), 'POST', {
+            data: contrib.serialize()
+        }).then(resp => {
+            contrib.unloadRecord();
+            this.store.pushPayload(resp);
+            let created = this.store.peekRecord('contributor', resp.data.id);
+            this.get('contributors').pushObject(created);
```
pattisdr: Can you explain these four lines, L186-189?

addon/mixins/node-actions.js
```diff
@@ -198,7 +170,6 @@ export default Ember.Mixin.create({
          */
         reorderContributors(contributor, newIndex) {
             contributor.set('index', newIndex);
-            //Node.save() or contributor.save(). Contributor.save() allows us to catch potential errors thrown.
             return contributor.save();
```
pattisdr: Store isn't modified with updated indices.

pattisdr:
Okay @chrisseto, I put in my two cents. Thanks for your work on this - I can correctly see when the request fails for add/update contribs which you could not do before this change.

Items:
- [ ] Update contributor does not work
- [ ] Store is incorrect when adding contributors - two records there, one with user id and one with correct contrib id
- [ ] Incorrect docstrings on updateContributor
- [ ] Store incorrect when removing contributors.
- [ ] Store isn't updated when reordering.

addon/models/node.js
chrisseto: Ha.. I know how to write code.

pattisdr:
@chrisseto - one thought for adding contributors, not calling contrib.unloadRecord() at all keeps one record in the store instead of having a duplicate, I do not know why.  Without this, there are two very similar records in the store, with the same ID but different _internalModels.

Also,  if there is an error adding a contributor, the contributor is still in the store even though the request was never sent to the server.  If you then try to add that contributor again, you get "ember.debug.js:19695 Uncaught Error: Assertion Failed: The id bgz3m-qjhye has already been used with another record of type preprint-service@model:contributor:."

PR #116, opened 2016-08-17
Allow same application to support cookie and token auth

abought:
## Ticket
https://openscience.atlassian.net/browse/PREP-118

Companion ember-preprints PR to follow.

# Purpose
Allow a single application to support both cookie and token based auth. Sample use case is to run preprints from within the OSF, as well as branded on different domains.

Imports are evaluated before code, so a new `OsfAgnosticAuth*` set of mixins was needed to support making auth configurable based on the value of `authorizationType` in `config/environment.js`.

## Future (for separate ticket)
For now, the token auth login button forces the user to go wherever the app registered to be the REDIRECT_URI (eg, the homepage). We may want more fine-grained redirects in the future (return the user to the page they were on when they clicked the login button)

I did experiment with REDIRECT_URIs that contained additional query params, but in each case CAS rejected these with an "invalid redirect_uri" error.

# TODO

## Login flow
- [x] Decision: how do branded app users get told they need to log in? (Answer: if they need to log in, they go to the login page)
- [x] Ensure that all components currently used by preprints work with both token and cookie auth (see #114)
- [x] Ensure that any hard-coded mixins (like login-route/controller) are selected appropriately based on the specified environment configuration when app is created. Eg, make the choice of mixin happen dynamically.

## Style and functionality
- [x] Navbar login button should match OSF
- [x] Navbar sign in button should match OSF
- [x] Provide a way for routes to check that the user is logged in, and test with...
  - [x] Cookie auth
  - [x] Token auth
- [x] For pages that require authentication, ember-simple-auth should send user to login page, and make sure they return to correct (ember) page when done
  - [x] For cookie auth
  - [x] For token auth (plan: use state parameter?)
- [x] Pages that do not require authentication should show logged-in-user in navbar, or show logged-out state
- [ ] Navbar login button should adjust dynamically when route changes, but still be a clickable link. (computed property cached?)

## Internal refactorings
- [x] Replace various authUrl / action based login usages with redirect behavior
- [x] Make sure ember-osf dummy app still works after changes (mixins, authUrl usages, login buttons etc)
- [ ] Authenticated AJAX should abstract away some of the extra headers manipulation silliness, then verify that manual request widgets (dropzone and fileManager) work...
  - [ ] For token auth
  - [ ] For cookie auth

# Notes for Reviewers
To be written.

## Tips for testing
- When switching between token and cookie auth, you may need to tell ember simple auth that what it thinks of as your login is no longer valid. Use `localStorage.clear()` at the console.
- When BACKEND=local(host), ember-osf is hardcoded to use personal access tokens. Edit `config/local.yml` accordingly.
  - *DO NOT* commit your personal access tokens to github!
  - To use real CAS:
    - Specify a `CLIENT_ID` in local.yml
    - Comment out [these lines](https://github.com/abought/ember-osf/blob/4f3fe0ff11ed6c23c41bea3b132f8b3516a17685/index.js#L48-L49)
    - Change the suffix of your [oauthUrl](https://github.com/abought/ember-osf/blob/4f3fe0ff11ed6c23c41bea3b132f8b3516a17685/index.js#L43-L43) from `/profile` to `/authorize` (similar to url used for production)
- This works on port 5000. On port 4200, using token auth, you may see errors unless `localhost:4200` is added to the API CORS ORIGIN whitelist:
https://github.com/abought/osf.io/blob/f0bb77351322667d240c05b7230e944eab12094f/api/base/settings/defaults.py#L112-L111

## Routes Added/Updated
- Application route change: do not block loading of application if user is not logged in. (let other auth route mixins do that)
- Login route changes to follow

### DIFF
--- a/addon/authenticators/osf-token.js
+++ b/addon/authenticators/osf-token.js
@@ -40,6 +40,11 @@ export default BaseAuthenticator.extend({
     },
     authenticate(accessToken) {
         // Adds the entire API user endpoint record to the session store as given
-        return this._test(accessToken);
+        let jqDeferred = this._test(accessToken);
+        return new Ember.RSVP.Promise((resolve, reject) => {
+            // TODO: Improve param capture
+            jqDeferred.done((value) => resolve(value));
+            jqDeferred.fail((reason) => reject(reason));
+        });
     }
 });
--- a/addon/mixins/osf-cookie-login-route.js
+++ b/addon/mixins/osf-cookie-login-route.js
@@ -1,7 +1,5 @@
 import Ember from 'ember';
 
-import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
-
 /**
  * @module ember
  * @submodule mixins
@@ -15,11 +13,14 @@ import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-
  * @extends Ember.Mixin
  * @uses ember-simple-auth/UnauthenticatedRouteMixin
  */
-export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
+export default Ember.Mixin.create({
     session: Ember.inject.service(),
     beforeModel() {
         // Determine whether the user is logged in by making a test request. This is quite a crude way of
         // determining whether the user has a cookie and should be improved in the future.
+
+        // TODO: Should this check for resolution of a promise?
+        this._super(...arguments);
 
         // Block transition until auth attempt resolves. If auth fails, let the page load normally.
         return this.get('session').authenticate('authenticator:osf-cookie')
--- a/addon/mixins/osf-cookie-login-route.js
+++ b/addon/mixins/osf-cookie-login-route.js
@@ -22,6 +22,7 @@ export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
         // determining whether the user has a cookie and should be improved in the future.
 
         // Block transition until auth attempt resolves. If auth fails, let the page load normally.
-        return this.get('session').authenticate('authenticator:osf-cookie').catch(err => console.log('Authentication failed: ', err));
+        return this.get('session').authenticate('authenticator:osf-cookie')
+            .catch(err => console.log('Authentication failed: ', err));
     }
 });
--- a/addon/mixins/osf-token-login-route.js
+++ b/addon/mixins/osf-token-login-route.js
@@ -1,5 +1,4 @@
 import Ember from 'ember';
-import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
 
 import config from 'ember-get-config';
 
@@ -18,7 +17,7 @@ import { getTokenFromHash } from 'ember-osf/utils/auth';
  * @class OsfTokenLoginRouteMixin
  * @extends Ember.Mixin
  */
-export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
+export default Ember.Mixin.create({
     session: Ember.inject.service(),
     beforeModel() {
         var accessToken;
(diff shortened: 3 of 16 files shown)


addon/authenticators/osf-token.js
```diff
@@ -40,6 +40,11 @@ export default BaseAuthenticator.extend({
     },
     authenticate(accessToken) {
         // Adds the entire API user endpoint record to the session store as given
-        return this._test(accessToken);
+        let jqDeferred = this._test(accessToken);
```
chrisseto: Doesn't ajax already return a promise/thennable object?

abought: Yes... but jquery promise semantics are slightly different from the ones ember is built on. This causes trouble when ember internals (like the beforeModel hook) try to deal with, eg, failure states.

(eg jquery promises lack a `.catch`)

It's quite ridiculous, actually. See, eg https://stackoverflow.com/questions/18808079/can-jquery-promises-be-chained-with-ember-rsvp-promises

addon/mixins/osf-cookie-login-route.js
```diff
@@ -15,13 +13,17 @@ import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-
  * @extends Ember.Mixin
  * @uses ember-simple-auth/UnauthenticatedRouteMixin
  */
-export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
```
chrisseto: What's the purpose of removing the` UnauthenticatedRouteMixin`?

addon/mixins/osf-token-login-route.js
```diff
@@ -33,6 +35,6 @@ export default Ember.Mixin.create(UnauthenticatedRouteMixin, {
         }
 
         return this.get('session').authenticate('authenticator:osf-token', accessToken)
-            .then(() => this.transitionTo('index'));
+            .catch(err => console.log('Authentication failed: ', err));
```
chrisseto: Will this swallow errors?

addon/utils/auth.js
```diff
@@ -15,22 +15,32 @@ import config from 'ember-get-config';
...
  */
-function getOAuthUrl() {
-    return `${config.OSF.oauthUrl}?response_type=token&scope=${config.OSF.scope}&client_id=${config.OSF.clientId}&redirect_uri=${encodeURI(config.OSF.redirectUri)}`;
+function getOAuthUrl(nextUri) {
+    // OAuth requires that redirect URI match what was registered, exactly. We may need a parameter to signify next
+    //   transition, if the user wants to return to ember at the point where they left off before needing to log in.
+    // For now, we will put this in the `state` parameter (always returned unchanged) and implement that functionality in ember later.
+    // To avoid abuse, the application should forcibly truncate state, eg make it relative to the application rootURL
+    //   (should not be possible to use the ember app as just an external redirect service)
+    let uri = `${config.OSF.oauthUrl}?response_type=token&scope=${config.OSF.scope}&client_id=${config.OSF.clientId}&redirect_uri=${encodeURI(config.OSF.redirectUri)}`;
```
chrisseto: Use `Ember.$.params`

addon/mixins/osf-cookie-login-route.js
abought: In the current cookie use case, there is no such thing as a "cookie login route" at all- it's simply mixed into the application route. (Your cookie login route is actually a CAS or flask app URL, not in ember at all)

This allows us to do things like checking for a valid cookie on first page load (so navbar shows correctly logged in state immediately).

From http://ember-simple-auth.com/api/classes/UnauthenticatedRouteMixin.html
> http://ember-simple-auth.com/api/classes/UnauthenticatedRouteMixin.html

addon/mixins/osf-token-login-route.js
abought: Mostly by design. Since the actual login form happens by going to an external URL, this mixin becomes much less responsible for logging someone in. Basically, the beforeModel hook is used to set up a page if someone is logged in.

If they're not logged in, that's ok too- it just means they see something a little different on the page. The underlying design assumption is that CAS controls the login page, and the ember mixin exists just to grab an access token and marking the page as logged in where appropriate.

Currently, CAS limits the redirectURI to a single specific URL, so this mixin could reasonably be used in, eg, just the index route of the site.

PR #117, opened 2016-08-19
Adapter Fixes

samchrisinger:
# Purpose
Fixes some of the adapter weirdness when working with Node contributors

## TODOS

- [X] Create contrib (Node relationship get's updated)
- [X] Delete contrib ...
- [X] Update contrib

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -29,6 +29,5 @@
   "white": false,
   "eqnull": true,
   "esnext": true,
-  "unused": true,
-  "requireSpacesInsideObjectBrackets": false
+  "unused": true
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -29,5 +29,6 @@
   "white": false,
   "eqnull": true,
   "esnext": true,
-  "unused": true
+  "unused": true,
+  "requireSpacesInsideObjectBrackets": false
 }
--- a/addon/adapters/contributor.js
+++ b/addon/adapters/contributor.js
@@ -19,10 +19,10 @@ export default OsfAdapter.extend({
                     node._internalModel.createSnapshot(),
                     'contributors'
                 );
-		// Needed for Ember Data to update the inverse record's (the node's) relationship
+                // Needed for Ember Data to update the inverse record's (the node's) relationship
                 return `${base}?embed=node`;
             } else {
-                // fail?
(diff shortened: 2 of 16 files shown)


pattisdr:
I appreciate your work on this Sam.

This fixes some of the items - before if you tried to add a contributor and the request failed you couldn't see that it had failed in ember-preprints except in the console.  So it would look like a contributor had been added and it would flash green for success but the gravatar would be missing, fullname, etc.

Items that are still broken - if you add contributors in quick succession, the email throttle will kick in and the request will fail.  The contributor is still in the store even though it never got added to the project.  Casey's PR will change the email throttle limit, but this is an example of where something could fail.
screen shot 2016-08-19 at 10 50 58 am

samchrisinger:
@pattisdr calling `Node.addContributor` will return a promise that resolves/rejects based off of the outcome of the save request (for the contributor). Internally ember-data will [automatically add the new contributor](https://github.com/emberjs/data/blob/c6d233b2ba1408f76d54f850bbe47bf8b2ae6982/addon/-private/system/store.js#L2491) to the Node's contributors relationship if the request succeeds. It won't if that request fails, and additionally we can respond to that failure.

pattisdr:
I am getting a 409 when trying to patch contributors.  The same contributor serializer is being used for both POST/PATCH requests for contributors.  POST needs user id, patch needs contributor id, which is not happening with current changes. 

Current PATCH request v2/nodes/node_id/contributors/contributor_id.  APIv2 flattens request, so user id under relationship probably overriding high-level id.  This is also not what the request should look like for PATCHing (we were already doing this, but as long as users/id was the contributor id, it worked).
```
{  
   "data":{  
      "id":"bgz3m-352qu",
      "attributes":{  
         "bibliographic":false
      },
      "type":"contributors",
      "relationships":{  
         "users":{  
            "data":{  
               "id":"352qu",
               "type":"users"
            }
         }
      }
   }
}
```

Response: {"errors":[{"detail":"The id you used in the URL, \"bgz3m-352qu\", does not match the id you used in the json body's id field, \"352qu\". The object \"bgz3m-352qu\" exists, otherwise you'd get a 404, so most likely you need to change the id field to match."}]}

pattisdr:
Patching contributors fixed - thanks @samchrisinger.

pattisdr:
@samchrisinger reordering contributors is broken now too, probably due to same reason as patching.

PR #118, opened 2016-08-19
[api-change] Improve taggable mixin by removing hard-coded path to taggable resource

abought:
## Ticket
https://openscience.atlassian.net/browse/PREP-55

❗ This PR is a dependency for https://github.com/CenterForOpenScience/ember-preprints/pull/46

⚠️ This is a backwards-incompatible change. It changes the call signature of actions on the TaggableMixin.

# Purpose
Supports tagging on pages where the thing to tag is not defined in the `model` hook of the route.

Changes how the `addATag` and `removeATag` mixins are used. Instead of passing `addATag=(action 'addATag')`, curry it so that the first argument is always implicitly there for any components (like `tags-widget`) that need it:

`addATag=(action 'addATag' model)`

This makes the mixin less tied to how variables are named on the route. (widgets only have to decide to do with tag text, not figure out what model to use)

# Notes for Reviewers
This will break any consuming apps that use the taggable mixin, until usages are updated. Sorry about that, but it's in a good cause...

Ultimately, we will probably want all usages of the mixin to work this way.

## Routes Added/Updated
Dummy app:
- File detail
- Node detail

Ember preprints:
- /submit

Audited other COS ember apps but none appeared to be using mixin yet.

### DIFF
--- a/addon/components/tags-widget/component.js
+++ b/addon/components/tags-widget/component.js
@@ -49,15 +49,20 @@ export default Ember.Component.extend({
                     this.send('removeATag', tag);
                 }
             });
-            // Populate the widget with the default tags passed in
-            let tags = this.get('tags');
-            this.$().importTags(tags.join(', '));
         });
+    }),
+
+    _doPopulateTags: Ember.observer('tags', function() {
+        // Rerender the list of tags whenever the node model changes. Useful if node.tags is not defined when page loads.
+        // Provide a default value in case tags weren't defined when component first rendered
+        this.$().importTags('');
+        let tags = this.get('tags') || [];
+        this.$().importTags(tags.join(', '));
     }),
 
     actions: {
         addATag(tag) {
-            //Calls a curried closure action which was provided the model
+            // Calls a curried closure action which was provided the model
             this.attrs.addATag(tag);
         },
         removeATag(tag) {
@@ -65,7 +70,7 @@ export default Ember.Component.extend({
             if (!tag) {
                 return false;
             }
-            this.sendAction('removeATag', tag);
+            this.attrs.removeATag(tag);
         }
     }
 });
--- a/addon/components/tags-widget/component.js
+++ b/addon/components/tags-widget/component.js
@@ -14,8 +14,8 @@ import layout from './template';
  *
  * ```handlebars
  * {{tags-widget
- *   addATag=(action 'addATag')
- *   removeATag=(action 'removeATag')
+ *   addATag=(action 'addATag' model)
+ *   removeATag=(action 'removeATag' model)
  *   tags=model.tags}}
  * ```
  * @class tags-widget
@@ -57,6 +57,7 @@ export default Ember.Component.extend({
 
     actions: {
         addATag(tag) {
+            //Calls a curried closure action which was provided the model
             this.attrs.addATag(tag);
         },
         removeATag(tag) {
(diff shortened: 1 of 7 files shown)


addon/components/tags-widget/component.js
```diff
@@ -49,22 +49,28 @@ export default Ember.Component.extend({
...
-            // Populate the widget with the default tags passed in
-            let tags = this.get('tags');
-            this.$().importTags(tags.join(', '));
         });
     }),
 
+    didRender() {
+        // Rerender the list of tags whenever the node model changes. Useful if node.tags is not defined when page loads.
+        // Provide a default value in case tags weren't defined when component first rendered
+        this.$().importTags('');
```
chrisseto: Is this line necessary, line 60 could be the same thing if tags is falsey.
Any reason not to hook this up to an observer on tags?

addon/components/tags-widget/component.js
```diff
@@ -49,22 +49,28 @@ export default Ember.Component.extend({
...
         addATag(tag) {
+            // Calls a curried closure action which was provided the model
             this.attrs.addATag(tag);
         },
         removeATag(tag) {
             // Don't try to delete a blank tag (would result in a server error)
             if (!tag) {
                 return false;
             }
-            this.sendAction('removeATag', tag);
```
chrisseto: Why change this? Wouldn't the ember way be to send actions up?

addon/mixins/taggable-mixin.js
```diff
@@ -18,28 +18,28 @@ export default Ember.Mixin.create({
...
-            var resource = this.get('model');
-            var currentTags = resource.get('tags').slice(0);
+        addATag(model, tag) {
+            var currentTags = model.get('tags').slice(0);
             Ember.A(currentTags);
             currentTags.pushObject(tag);
-            resource.set('tags', currentTags);
-            return resource.save();
+            model.set('tags', currentTags);
+            return model.save();
```
chrisseto: Does `model.get('tags').pushObject(tag)` not work here?

addon/components/tags-widget/component.js
abought: As discussed last week- when this was hooked up to an observer, it caused unit tests to fail.

addon/components/tags-widget/component.js
abought: We're calling it this way to take advantage of a quirk of closure actions. Calling `.send` or `.sendAction` is a fire-and-forget process, but there are times we want our components to respond to the result of an action. (eg to clean up internal state for the component if a promise/save fails) 

In the future we may choose to respond to the result of a failed promise in deciding how to update a context-specific save failure message field in the component, for example.

https://alisdair.mcdiarmid.org/ember-closure-actions-have-return-values/

addon/mixins/taggable-mixin.js
abought: Actually, not 100% sure as to current state, with the adapter changes merged the last few days. 

These lines of code were in the original, and at the time- whether due to adapter customization or something else- the save literally didn't work at all. (I recall being particularly weirded out by line 26)

addon/components/tags-widget/component.js
abought: As for consolidation, it looks like the reset (line 58) can indeed be consolidated with populating the new tags (line 60). I'll push that with any other changes requested when review is done.

https://github.com/xoxco/jQuery-Tags-Input/blob/ae31b175fbfd0a0476822182ef52e76c2629e9c3/src/jquery.tagsinput.js#L171

PR #123, opened 2016-08-25
Add hideSearch component param to osf-navbar

hmoco:

### DIFF
--- a/addon/components/osf-navbar/component.js
+++ b/addon/components/osf-navbar/component.js
@@ -12,12 +12,24 @@ import { getAuthUrl } from 'ember-osf/utils/auth';
 /**
  * Display the OSF navbar
  * @class osf-navbar
+ * Sample usage:
+ * ```handlebars
+ * {{osf-navbar
+ *   hideSearch=true
+ *  }}
+ * ```
  */
 export default Ember.Component.extend({
     layout,
     session: Ember.inject.service(),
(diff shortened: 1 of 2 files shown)


abought:
Can you provide documentation for this parameter, and a default value in the controller in case the user doesn't specify one?  Future users of the component should be able to figure out what it does and how to use it/ where it comes from.

This is one area where ember components are kind of weird- it's not always easy to reuse them across different projects. Working from example gets harder when the examples are in many different github repos!

Sample usage example: https://github.com/abought/ember-osf/blob/ea1462dfd277155da6a3d803075fd7e375bb8638/addon/components/eosf-project-nav/component.js#L15-L19
YUIDoc sample format for an individual property, https://github.com/abought/ember-osf/blob/368ecb93941976f91d67a8317f017fdb84d01fcd/addon/components/tags-widget/component.js#L32-L32

The result of those docstrings is something like this (we use YUIDoc format):
https://abought.github.io/ember-osf/classes/tags-widget.html

hmoco:
@abought let me know when you can merge this

abought:
Also- I can fix this later, but it seems like some information got lost in a merge commit as we both updated the sample usage.

https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/components/osf-navbar/component.js#L10-L17

abought:
Looks good. Will merge when tests passing.

PR #125, opened 2016-08-25
Add the ability to pass along sendEmail=False when adding a contributor

erinspace:
## Ticket
Vaguely related to Preprints Ticket - https://openscience.atlassian.net/browse/PREP-130

# Purpose
Preprints needed the ability to send an email to preprint authors after the whole preprint process has been completed, which does not happen until after a normal project has been created and contributors added to it. So, we needed a way to pass along OSF API v2's query parameter `send_email=False` when creating a contributor from Ember OSF through the preprints interface! 

Then, the email can be sent after the whole preprint process is completed, and users unfamiliar with the OSF can receive an email with language about the preprint and not an unanticipated "project."

# Notes for Reviewers
- I am very new to ember, so style, testing, really any improvement pointers welcome!
- I also may very well have broken other places within ember OSF that use the addContributor method (tho I tried to cover the bases). Is there a way to add a default parameter? This has been locally tested with ember-preprints, but have not done functional testing on ember-osf. 
- Does SHARE call this method ever? In its curation interface maybe?

## Routes Added/Updated
- Adapter for contributor to pass along `?send_email=false` when adding a contributor with that set.
- send along the query parameter in the node-actions mixin
- `sendEmail` added to the contributor model
- pass along `sendEmail` in the node model
- also in the contrib-manager component test

### DIFF
--- a/addon/adapters/contributor.js
+++ b/addon/adapters/contributor.js
@@ -7,7 +7,7 @@ export default OsfAdapter.extend({
             var sendEmail = true;
             if (snapshot) {
                 nodeId = snapshot.record.get('nodeId');
-                sendEmail = snapshot.record.get('sendEmail')
+                sendEmail = snapshot.record.get('sendEmail');
             } else {
                 nodeId = id.split('-').shift();
             }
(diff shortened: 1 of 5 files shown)


addon/adapters/contributor.js
```diff
@@ -21,8 +23,14 @@ export default OsfAdapter.extend({
                     return `${base}${id.split('-').pop()}/`;
                 }
 
+                var requestUrl = `${base}?embed=node`;
+
+                if (!sendEmail) {
+                    requestUrl += `&send_email=false`;
+                }
+
                 // Needed for Ember Data to update the inverse record's (the node's) relationship
```
samchrisinger: This comment belongs with the `?embed=node`

PR #129, opened 2016-08-29
Feature/Manually update store as reordering contributors

pattisdr:
❗ Needed for https://github.com/CenterForOpenScience/ember-preprints/pull/113
# Purpose

APIv2 reordering contributor only takes in contributor you are modifying and that contributor's new index.  Ember store was only being updated with that particular contributor's index - no other contributors were changing which could lead to this case:

screen shot 2016-08-29 at 7 11 50 am

So one index is correct, the other is incorrect. 

# Changes
Manually updates remaining contributors in the store with correct index.  Thanks to @samchrisinger for talking through this with me.

# Notes for Reviewers


pattisdr:
Failing test has to do with file renderer - nothing to do with this PR.

abought:
> Failing test has to do with file renderer - nothing to do with this PR.

Failing tests should be addressed in #130 . :)

pattisdr:
Thank you @abought!

PR #132, opened 2016-08-31
Feature/iframe change height dynamically for text files

pattisdr:
# Ticket 
https://openscience.atlassian.net/browse/PREP-146

# Purpose
Makes height of iframe same height as txt file.  More work needed for rendering pdfs, ppts, .docx, etc.

# Notes for Reviewers


pattisdr:
Maybe not such a good idea to render entire txt file on page, since it bumps file selector to bottom. Still have unresolved issue to get pdfs to render properly.  Closing for now.

PR #133, opened 2016-09-08
Add allowfullscreen as default in file-renderer [PREP-147]

caneruguz:
## Ticket
https://openscience.atlassian.net/browse/PREP-147

# Purpose
It looks like the mfr options do not include allowing full screen as default. This is useful for preprints and I think it would be a good default to have. 

# Notes for Reviewers
I don't know if the optionality of this needs to be worked out more. I did add it as a variable and did not hard code it in but it's the default. 

## Routes Added/Updated
NA

### DIFF
--- a/addon/components/file-renderer/component.js
+++ b/addon/components/file-renderer/component.js
@@ -23,7 +23,7 @@ export default Ember.Component.extend({
     download: null,
     width: '100%',
     height: '100%',
-    allowfullscreen : true,
+    allowfullscreen: true,
     mfrUrl: Ember.computed('download', function() {
         var base = config.OSF.renderUrl;
         var download = this.get('download');
--- a/addon/components/file-renderer/component.js
+++ b/addon/components/file-renderer/component.js
@@ -14,7 +14,7 @@ import config from 'ember-get-config';
  * ```handlebars
  * {{file-renderer
  *   download=model.links.download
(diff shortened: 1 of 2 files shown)


pattisdr:
In the component.js file, under the Sample usage of the file-renderer component, I would add allowfullscreen=allowfullscreen. 

https://github.com/caneruguz/ember-osf/blob/fd534e8bd99374d95ca984a8ee52a533987fea76/addon/components/file-renderer/component.js#L17

pattisdr:
Any reason we would need some of these options `mozallowfullscreen` or `webkitallowfullscreen`?

pattisdr:
@caneruguz pass done, one clarifying question and a request to add addfullscreen option to the sample usage.

caneruguz:
> Any reason we would need some of these options mozallowfullscreen or webkitallowfullscreen?
Good question, it works in Safari without webkitallowfullscreen for instance. I'll find this out.

caneruguz:
Apparently the browser specific ones are deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-mozallowfullscreen

pattisdr:
Ok - just this  https://github.com/CenterForOpenScience/ember-osf/pull/133#issuecomment-245687034 and I am good here

PR #134, opened 2016-09-13
[Ready to Review] Feature/Embed Root and Parent on User Nodes

pattisdr:
## Ticket

Needed for https://github.com/CenterForOpenScience/ember-preprints/pull/141

Selecting existing project for preprints - need to get format to resemble dashboard format.  For example,  Project / ... / Private / Component, instead of Component".  

# Purpose

Changes 
- [x] Initial query params were not getting passed into subsequent recursive calls on loadAll
- [x] Allow user to embed both root and parent when fetching userNodes 
- [x]  Allow embedding of data with errors (for example fetching private parent)
- [x] Allow discernment of whether not request failed with 403 or 404. (This is so I can distinguish whether a parent doesn't exist or it's private in preprints and can be formatted appropriately. (@hmoco's ember-preprints changes take care of this so you don't need to distinguish between errors)

# Notes for Reviewers

### DIFF
--- a/addon/adapters/user.js
+++ b/addon/adapters/user.js
@@ -1,4 +1,24 @@
 import OsfAdapter from './osf-adapter';
+import Ember from 'ember';
 
 export default OsfAdapter.extend({
+    findHasMany(store, snapshot, url, relationship) {
+        var id = snapshot.id;
+        var type = snapshot.modelName;
+
+        url = this.urlPrefix(url, this.buildURL(type, id, null, 'findHasMany'));
+
+        // If fetching user nodes, will embed root and parent.
+        if (relationship.type === 'node') {
+            var embedParams = 'embed=parent&embed=root';
+            if (snapshot.record.get('query-params')) {
+                var queryParams = Ember.$.param(snapshot.record.get('query-params'));
+                url += '?' + queryParams;
+                url += '&' + embedParams;
+            } else {
+                url += '?' + embedParams;
+            }
+        }
+        return this.ajax(url, 'GET');
+    }
 });
--- a/addon/adapters/user.js
+++ b/addon/adapters/user.js
@@ -10,13 +10,9 @@ export default OsfAdapter.extend({
 
         // If fetching user nodes, will embed root and parent.
         if (relationship.type === 'node') {
-            var embedParams = 'embed=parent&embed=root';
+            url +='?embed=parent&embed=root';
             if (snapshot.record.get('query-params')) {
-                var queryParams = Ember.$.param(snapshot.record.get('query-params'));
-                url += '?' + queryParams;
-                url += '&' + embedParams;
-            } else {
-                url += '?' + embedParams;
+                url += '&' + Ember.$.param(snapshot.record.get('query-params'));
             }
         }
         return this.ajax(url, 'GET');
--- a/addon/adapters/user.js
+++ b/addon/adapters/user.js
@@ -10,7 +10,7 @@ export default OsfAdapter.extend({
 
         // If fetching user nodes, will embed root and parent.
         if (relationship.type === 'node') {
-            url +='?embed=parent&embed=root';
+            url += '?embed=parent&embed=root';
             if (snapshot.record.get('query-params')) {
                 url += '&' + Ember.$.param(snapshot.record.get('query-params'));
             }
--- a/addon/adapters/user.js
+++ b/addon/adapters/user.js
@@ -6,7 +6,7 @@ export default OsfAdapter.extend({
         var id = snapshot.id;
         var type = snapshot.modelName;
 
-        url = this.urlPrefix(url, this.buildURL(type, id, null, 'findHasMany'));
+        url = this.urlPrefix(url, this.buildURL(type, id, snapshot, 'findHasMany'));
 
         // If fetching user nodes, will embed root and parent.
         if (relationship.type === 'node') {
--- a/addon/models/osf-model.js
+++ b/addon/models/osf-model.js
@@ -17,6 +17,7 @@ import HasManyQuery from 'ember-data-has-many-query';
 export default DS.Model.extend(HasManyQuery.ModelMixin, {
     links: DS.attr('links'),
     embeds: DS.attr('embed'),
+    errors: DS.attr(),
 
     relationshipLinks: Ember.computed.alias('links.relationships'),
     _dirtyRelationships: null,
--- a/addon/models/osf-model.js
+++ b/addon/models/osf-model.js
@@ -17,7 +17,6 @@ import HasManyQuery from 'ember-data-has-many-query';
 export default DS.Model.extend(HasManyQuery.ModelMixin, {
     links: DS.attr('links'),
     embeds: DS.attr('embed'),
-    errors: DS.attr(),
 
     relationshipLinks: Ember.computed.alias('links.relationships'),
     _dirtyRelationships: null,
--- a/addon/utils/load-relationship.js
+++ b/addon/utils/load-relationship.js
@@ -15,10 +15,9 @@ export default function loadAll(model, relationship, dest, options = {}) {
         var pageSize = results.meta.pagination.per_page;
         var remaining = total - (page * pageSize);
         if (remaining > 0) {
-            return loadAll(model, relationship, dest, {
-                'page[size]': pageSize,
-                page: page + 1
-            });
+            query.page = page + 1;
+            query['page[size]'] = pageSize;
+            return loadAll(model, relationship, dest, query);
         }
     });
 }
--- a/addon/utils/load-relationship.js
+++ b/addon/utils/load-relationship.js
@@ -7,7 +7,7 @@ export default function loadAll(model, relationship, dest, options = {}) {
         page: page
     };
     query = Ember.merge(query, options || {});
-    model.set('query-params', query);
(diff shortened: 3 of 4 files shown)


addon/models/osf-model.js
```diff
@@ -17,6 +17,7 @@ import HasManyQuery from 'ember-data-has-many-query';
 export default DS.Model.extend(HasManyQuery.ModelMixin, {
     links: DS.attr('links'),
     embeds: DS.attr('embed'),
+    errors: DS.attr(),
```
abought: Shouldn't every `DS.Model` already have an errors property populated by default?

http://emberjs.com/api/data/classes/DS.Model.html#property_errors

pattisdr: Good to know

pattisdr: What about getting errors on relationships?  If I fetch node.parent and there is no parent, or I don't have permission to view the parent?

abought: This is kind of an awful one, because by default typing `node.parent` would mess up the template. Ember assumes the relationship corresponds to a request that will work.

Here's how I hacked around it in the project navbar, which has a "parent is private" indicator... I checked that the parent relationship existed:
https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/components/eosf-project-nav/component.js#L49

And then a computed property that returned false if the parent couldn't be successfully fetched...
https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/components/eosf-project-nav/component.js#L55

There might be a more elegant way to go about it, so alternate ideas very welcome.

pattisdr: That is helpful, thank you.  Another problem I have is that I need to know which error it received when it failed.  I am supposed to make project selector match OSF dashboard.

Dashboard tries to fetch parent. If it doesn't exist, then it doesn't show up in tree.  If it doesn't exist, but 403's, then it shows up as "private".  For example, Grandparent / Private/ Child.  The parent was private and it didn't show up. But to be able to distinguish, I need to know if it 404's or 403's.

abought: When you catch the error on an attempted request, it should provide the status code for the failed request, correct?

pattisdr: These are embedded parents and roots.  nodes/node_id/?embed=parent.   The embedded parent has no indication what the problem was, just content=null.  It does make a request afterwards to fetch the parent itself which 403's.   I wanted to know if there was any way in the embedded context to preserve this information.  

@hmoco is going to take over looking into this.  Thanks for your helpful comments.

addon/adapters/user.js
```diff
@@ -1,4 +1,24 @@
...
+
+        // If fetching user nodes, will embed root and parent.
+        if (relationship.type === 'node') {
+            var embedParams = 'embed=parent&embed=root';
+            if (snapshot.record.get('query-params')) {
+                var queryParams = Ember.$.param(snapshot.record.get('query-params'));
+                url += '?' + queryParams;
+                url += '&' + embedParams;
+            } else {
+                url += '?' + embedParams;
```
hmoco: Minor thing, could this logic be simplified to ` embedParams = '?embed=parent&embed=root'` and then adding `& + queryParams ` if they are defined?

addon/utils/load-relationship.js
```diff
@@ -15,10 +16,9 @@ export default function loadAll(model, relationship, dest, options = {}) {
         var pageSize = results.meta.pagination.per_page;
         var remaining = total - (page * pageSize);
         if (remaining > 0) {
-            return loadAll(model, relationship, dest, {
-                'page[size]': pageSize,
-                page: page + 1
-            });
+            query.page = page + 1;
+            query['page[size]'] = pageSize;
+            return loadAll(model, relationship, dest, query);
```
hmoco: 👍

hmoco:
@caneruguz, @pattisdr, this is good to go! One minor change if you have time, but not needed.

addon/adapters/user.js
pattisdr: good idea :)

addon/adapters/user.js
```diff
@@ -1,4 +1,20 @@
 import OsfAdapter from './osf-adapter';
+import Ember from 'ember';
 
 export default OsfAdapter.extend({
+    findHasMany(store, snapshot, url, relationship) {
```
abought: From a forward-maintainability standpoint, this override seems to omit certain features in newer versions of ember that might be relevant (and even helpful for your specific use case).

http://emberjs.com/blog/2016/06/09/ember-data-2-6-released.html#toc_code-ds-improved-ajax-code-a-href-https-github-com-emberjs-data-pull-3099-3099-a

https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L630

addon/utils/load-relationship.js
```diff
@@ -7,6 +7,7 @@ export default function loadAll(model, relationship, dest, options = {}) {
         page: page
     };
     query = Ember.merge(query, options || {});
+    Ember.set(model, 'query-params', query);
```
abought: Is this a special field on the model that we control? Do we need to clear the field in between requests?

The two scenarios I'm considering here are:
- The same model gets interrogated later and possibly remembers a previous setting, thus not starting fresh, and
-  In some cases, setting extra fields on a model to store information can cause the model to be marked as dirty, even if the fields have no real impact on a request

addon/adapters/user.js
pattisdr: We are using Ember-data 2.3.0 in Ember-OSF and ember-data 2.7.0 in Ember-preprints.  To make use of these features, would Ember-OSF need to be updated to 2.7?

The reason I am doing this -  ember doesn't let you pass query params to relationships.  so we are using the `ember-data-has-many-query` addon. To use this addon, query params are passed in a dictionary.   The only problem is, for embeds, the keys are the same, and I need to embed more than one relationship. I am open to a better way of doing this - but I think I am going to need to build the url myself.

addon/utils/load-relationship.js
pattisdr: I agree that this could potentially be problematic.  I am building the URL myself in findHasMany to embed both `parent` and `root`, but I still needed to pass on any other query params when I build the url.  I put them on the model so I would have them in findHasMany.  Maybe there is a better way to pass along this information.

addon/adapters/user.js
abought: Per version differences: long-term, I'd encourage us to move `ember-osf` to Ember 2.8 (which was released earlier this month, and is currently a candidate LTS build with a more stable support commitment).

However, I also recognize that upgrading ember versions is not a trivial task; it's a separate ticket in itself. We should probably create such a ticket (and note this as an area to change). But so long as we're pinned to an old ember version this approach is fine for now.

abought: So to clarify, I assume the key problem is that we use two embeds?
`?embed=parent&embed=root'`

... and thus it doesn't work with a query object of form `{embed: 'parent', embed:'root'}` ?

Realistically, this seems like an API limitation that we'll have to keep hacking around in the future- this won't be the last time we need more than one embed in a request. Sometimes when a single parameter has multiple values, I've see APIs accept a comma-separated list of values: `embed=parent,root`. That seems in keeping with the JSONAPI spec, eg 
http://jsonapi.org/format/#fetching-includes

@brianjgeiger , do you have any thoughts on the best way to support multiple values for a single query param in the API going forward?

 I've [confirmed](https://staging-api.osf.io/v2/users/me/nodes/?embed=parent,root) that our API doesn't support that for the embed parameter at present.

caneruguz:
So @abought and @pattisdr , does this require additional development before it can be merged in? What is PR specific action that we need to take here without considerations for overall development and new tickets?

pattisdr:
@abought just suggested a solution to me, I need a few minutes to see if it works.

addon/adapters/user.js
brianjgeiger: Unless I have to, I'm going to try to avoid loading the entire context for this in order to completely answer the question, but if I need to, we can have a discussion. That being said, I am okay with either: 1) allowing `?embed=root,parent`; or 2) allowing headers that specify what is embedded (as long as conflicts either `Conflict` or do the superset, in this case). If either of those solutions make this easier, go for it. If more understanding is required, let's set aside some time.

abought: Apologies for the noise. Some further experimentation suggested that we may be able to use the following after all:
`return this.store.query('account', {q:[1,2,3,4]});`

to get a url of the form needed, eg

`https://staging-metadata.osf.io/v1/id/collections/experimenter.accounts/_search?q=1&q=2&q=3&q=4`

This would let Dawn do exactly what she needs in a simpler fashion. (sorry for the noise- ember specs, docs, and various methods aren't always consistent with handling of arrays) She'll try it out as a direct path and report back.

If it doesn't work out, the path forward is to merge this now, but figure out API in the future before embeds come up again.

abought:
We've investigated the low-hanging fruit, but the most appealing option did not work consistently across ember versions. 

Since we'll eventually need a reliable way to provide multiple embeds in the future, Dawn will file an improvement ticket for a later date... but we'll merge this as is to move this work forward. (if it's not a quick fix, it's not worth getting this bogged down)

caneruguz:
Thank you @abought and @pattisdr for working through this.

PR #143, opened 2016-10-26
Moving out Navbar Auth Dropdown into its own component [PREP-217]

binoculars:
# Purpose

Needed for ember-preprints branded navbar.
# Notes for Reviewers

Should be no visible functional changes on the osf-navbar

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "docs": "yuidoc",
     "docs-deploy": "yuidoc && ghp-import -r origin -p -b gh-pages docs/",
     "start": "ember server",
-    "test": "./node_modules/jscs/bin/jscs . && ember test"
+    "check-style": "./node_modules/jscs/bin/jscs .",
+    "test": "npm run-script check-style && ember test"
   },
   "repository": "",
   "engines": {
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -11,8 +11,7 @@
     "docs": "yuidoc",
     "docs-deploy": "yuidoc && ghp-import -r origin -p -b gh-pages docs/",
     "start": "ember server",
-    "test": "ember test",
-    "check-style": "./node_modules/jscs/bin/jscs ."
+    "test": "./node_modules/jscs/bin/jscs . && ember test"
```
caneruguz: I would leave these separate, checking style is not same as testing and there may be reasons to do these tasks individually.

binoculars: You can still run them separately, but travis enforces style checking, so it's important to have the same command available locally to ensure tests are passing. I've seen multiple instances where myself and others have:  
1. pushed a commit
2. had to wait for travis to run
3. the build fails because check-style wasn't run
4. fix the style errors and run check-style again locally
5. push another commit and wait for it to pass in travis

We could avoid those 5 steps by enforcing the check-style up front. If you don't want to run check-style right away, you can just run ember test. However, `npm test` should run all tests that you would expect to have in the CI environment.

caneruguz: These five steps are happening because we are forgetting to run things. This is only useful if  you are running npm test and forgetting to check the style. Adding it in is fine but please keep the check-style as a separate option without having to write it in full.

PR #151, opened 2016-11-15
Add ability to have a redirect url after logout [PREP-253]

binoculars:
## Ticket
https://openscience.atlassian.net/browse/PREP-253

# Purpose
Needed for branded preprints redirects

# Notes for Reviewers
* redirect_url is appended to the logout page as a query param, if the value is set on the component. If not, it goes to the default logout page
* Add trailing slash after login for the login uri

## Routes Added/Updated
N/A

## Related
CenterForOpenScience/ember-preprints#190

### DIFF
--- a/addon/components/navbar-auth-dropdown/component.js
+++ b/addon/components/navbar-auth-dropdown/component.js
@@ -17,7 +17,7 @@ import config from 'ember-get-config';
  *   redirectUrl=redirectUrl}}
  * ```
  *
- * @class osf-navbar
+ * @class navbar-auth-dropdown
  */
 export default Ember.Component.extend({
     layout,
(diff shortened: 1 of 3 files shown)


addon/components/navbar-auth-dropdown/component.js
```diff
@@ -13,7 +13,8 @@ import config from 'ember-get-config';
  * Sample usage:
  * ```handlebars
  * {{navbar-auth-dropdown
- *   loginAction=loginAction}}
+ *   loginAction=loginAction
+ *   redirectUrl=redirectUrl}}
  * ```
  *
  * @class osf-navbar
```
abought: Tiny note: the docstring may need to be updated here to reflect the component name. Otherwise YUIDoc might not build it to the correct docs page.

PR #152, opened 2016-11-15
License Picker Ember widget (to be used for preprints) [EOSF-155]

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-155

# Purpose
Create a license picker in ember with feature parity with the current non-ember license widget used for nodes that is currently using a v1 route. Also needs to support preprints and be embeddable in the create/edit preprint Basics section. Needs to take in a list of available preprints as well, where order matters (top is default).
(description is wip as well)
(not sure what those other commits are, will rebase)


addon/components/license-picker/component.js
```diff
@@ -0,0 +1,22 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
```
abought: Please make sure to see a YUIDoc docstring, including sample usage, for all ember-osf components. Otherwise this component will be invisible to the [documentation](http://centerforopenscience.github.io/ember-osf/).

See other components in this repo for an example and let me know if I can help. To verify this works correctly, you can run `npm run docs` with YUIDoc installed.

hmoco: I plan on doing so! when I figure out what will actually go in this.

addon/components/license-list/component.js
```diff
@@ -0,0 +1,75 @@
...
+                categories.push({
+                    title: key,
+                    licenses: cat[key]
+                });
+            }
+        }
+        if (hasOther) {
+            categories.push(hasOther);
+        }
+        this.set('categories', categories);
```
pattisdr: Question about one of the categories in this list that we're returning from the API -  the "Academic Free License (AFL) 3.0" is not in the OSF project overview page dropdown. Do we need to add it there?

addon/components/license-list/component.js
```diff
@@ -0,0 +1,75 @@
+import Ember from 'ember';
+import layout from './template';
+
+const defaultCategories = {
+    Content: name => name.indexOf('CC') !== -1,
+    'Code - Permissive': name => name.indexOf('MIT') !== -1 || name.indexOf('Apache') !== -1 || name.indexOf('BSD') !== -1,
+    'Code - Copyleft': name => name.indexOf('GNU') !== -1 && name.indexOf('Lesser') === -1
```
pattisdr: Use `includes` instead of `indexOf` to see if a string is in another string, ex. `name.includes('CC')`

addon/components/license-list/component.js
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+        if (hasOther) {
+            categories.push(hasOther);
+        }
+        this.set('categories', categories);
+    }),
+    actions: {
+        select() {
+            let license = this.get('store').peekRecord('license', this.$('select')[0].value);
```
pattisdr: prefer not accessing the license id this way - it could conflict with other items on the page.  If you add an argument to the select action, and `value=target.value` instead of  `this` in the license-list template, you should have your id. 
```js
 select(target) {
     let license = this.get('store').peekRecord('license', target);
     ...
}
```

addon/components/license-list/template.hbs
```diff
@@ -0,0 +1,22 @@
+
+<select class="form-control" onchange={{action 'select' this}}>
```
pattisdr: Instead, 
```js
onchange={{action 'select' value='target.value'}}
```

addon/components/license-picker/component.js
```diff
@@ -0,0 +1,78 @@
...
+        if (!text) {
+            return '';
+        }
+        text = text.replace(/({{year}})/g, this.get('year') || '');
+        text = text.replace(/({{copyrightHolders}})/g, this.get('copyrightHolders') || '');
+        return text;
+    }),
+    licenseEdited: Ember.observer('copyrightHolders', 'nodeLicense', 'year', function() {
+        Ember.run.debounce(this, function() {
+            if (this.get('autosave')) {
```
pattisdr: Where would `autosave` be set? Would this be a property passed in to the component? It should have a default value on the component as well.

addon/components/license-picker/component.js
```diff
@@ -0,0 +1,78 @@
...
+                this.get('actions.save').bind(this)();
```
pattisdr: Also, can you help me understand what this line does?

pattisdr:
Also, preprint-provider model needs `licensesAcceptable`.

addon/components/license-picker/component.js
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            this.set('year', this.get('currentValues.year'));
+        }
+        if (this.get('currentValues.copyrightHolders')) {
+            this.set('copyrightHolders', this.get('currentValues.copyrightHolders'));
+        }
+    },
+    _setNodeLicense: Ember.observer('licensesAvailable', function() {
+        if (!this.get('currentValues.licenseType')) {
+            this.set('nodeLicense', this.get('licensesAvailable.firstObject'));
```
hmoco: That is because providers will have default licenses that need to be preselected. Here they would be the first in the array.

addon/components/license-picker/component.js
hmoco: Autosave is set on the edit and add preprint, so that no button click is used to indicate a change of state.

addon/components/license-picker/component.js
hmoco: If changes are made to the form, and autosave is true, this component should fire the save action normally used by the Save button.

addon/components/license-list/component.js
hmoco: @icereval any insight?

addon/models/node.js
```diff
@@ -38,6 +38,7 @@ export default OsfModel.extend(FileItemMixin, {
     dateCreated: DS.attr('date'),
     dateModified: DS.attr('date'),
 
+    nodeLicense: DS.attr(),
```
hmoco: Yes, but there they are called licenseRecord, since that makes more sense in the general concept of a license, when attached to something that is not a node.

addon/serializers/node.js
```diff
@@ -1,4 +1,26 @@
...
 
 export default OsfSerializer.extend({
+    serialize(snapshot) {
+        // Normal OSF serializer strips out relationships. We need to add back primaryFile/node/provider for this endpoint
+        const res = this._super(...arguments);
+        res.data.relationships = {};
+        let hasRelation = false;
+        for (var rel in snapshot.record._dirtyRelationships) {
+            let relationship = Ember.String.underscore(rel);
+            if (relationship.includes('license')) {
```
pattisdr: Good catch on this one - I completely forgot we were stripping relationships out.

brianjgeiger:
@hmoco Could you fix the failing test(s)?

PR #155, opened 2016-11-23
Feature/Citation Widget [PREP-139]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/PREP-139

# Purpose

Reusable ember-osf citation widget component.  Currently only provides three citations: MLA, Chicago, and APA.  Does not provide dropdown to fetch additional citations.

Turns out is very similar to @hmoco's work here: https://github.com/CenterForOpenScience/ember-preprints/pull/144, since I am just making ajax requests for now.  Main difference is that I have pulled this to ember-osf.

# Notes for Reviewers
I am making ajax requests to fetch the node's citation in APA, MLA, and Chicago formats, instead of the Ember way. This is because the API endpoints are unusual.   `nodes/node_id/citation` is a 'node-citation' type, but `nodes/node_id/citation/citation_id/` is a 'styled-citations' type.   Node citation detail endpoint and node citations styles detail endpoints don't seem to be related.  I recommend future work for coming up with a more Ember way to fetch these citations (override findBelongsTo, maybe?), or perhaps changing how the endpoints are structured API side.


pattisdr:
hm why are ember-osf tests failing now?

brianjgeiger:
@pattisdr Tests are failing on this. Could you merge with develop and see if they're better? Probably Yarn fixes it.

pattisdr:
Yes it should fix it I had been waiting for @binoculars's work to be finished.

brianjgeiger:
I think there's going to have to be one more fix, so hold up before you merge. https://openscience.atlassian.net/browse/EOSF-463

brianjgeiger:
Well, now we're only failing 2 tests.

PR #159, opened 2016-12-05
[PREP-278] CopyrightHolders should always return a string to the controller

hmoco:
## Purpose
From the perspective of the component, copyrightHolders it should be only a string, never an array. This helps fix licenseChanged not always being accurate on preprints edit.

### DIFF
--- a/addon/components/license-picker/component.js
+++ b/addon/components/license-picker/component.js
@@ -79,7 +79,7 @@ export default Ember.Component.extend({
             let values = {
                 licenseType: this.get('nodeLicense'),
                 year: this.get('year'),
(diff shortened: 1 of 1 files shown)


brianjgeiger:
But the API is still expecting a list, yes? And the model is still expecting a list?

hmoco:
Yep. But the component doesn't handle interactions with the API. Those are handled by the controller consuming this component, while it passes back what the fields are currently populated with.

brianjgeiger:
Okay, so it'll still interact properly, just further up the chain?

hmoco:
Exactly!

pattisdr:
Changes look good to me.

PR #165, opened 2017-01-10
Feature/Verify results.meta in load-relationship [EOSF-400]

pattisdr:
## Ticket

Needed for https://openscience.atlassian.net/browse/EOSF-400

# Purpose

Verify that results.meta is present in the response before accessing properties of results.meta. 

# Notes for Reviewers

This is useful for preprints tests. Many tests fail because `Cannot read property 'pagination' of null`.  being called in load-relationship when there is no `results.meta`.   Lots of observers fire which causes load-relationship to get called often in tests, even though it is unrelated to the test itself. For example, creating nodes causes the `getContributors` observer to fire which causes the ember-osf `loadRelationship` to run. Because there are not contributors set on the node, it fails.  If I add contributors to the nodes in the test, then I am getting `calling set on destroyed object` errors.

While I don't like modifying code just for tests, this really helps testing, and I don't think it hurts to check that results.meta is present before proceeding.


pattisdr:
Tests passing locally? hm.

addon/utils/load-relationship.js
```diff
@@ -11,14 +11,15 @@ export default function loadAll(model, relationship, dest, options = {}) {
...
-        var total = results.meta.pagination.total;
-        var pageSize = results.meta.pagination.per_page;
-        var remaining = total - (page * pageSize);
-        if (remaining > 0) {
-            query.page = page + 1;
-            query['page[size]'] = pageSize;
-            return loadAll(model, relationship, dest, query);
+        if (results.meta) {
+            var total = results.meta.pagination.total;
+            var pageSize = results.meta.pagination.per_page;
```
abought: It seems reasonable that a user depaginating the API might want to start by fetching more than the default minimum per page, by passing in the `query` initially.

As written, it seems like the "get next page after first" call would overwrite the original query page size. Eg the user could fetch 50 results the first page (because that's what they said to get), but only 10 thereafter (or whatever the API says to get next). 

Is overriding the user-provided `query` the intended behavior? Or do we want this to be a default value if none was specified explicitly?

brianjgeiger:
Page size changes through the API pass through to future pages. So https://api.osf.io/v2/nodes/?page[size]=5 will return pagination links like https://api.osf.io/v2/nodes/?page=2&page%5Bsize%5D=5

addon/utils/load-relationship.js
brianjgeiger: Page size changes through the API pass through to future pages. So https://api.osf.io/v2/nodes/?page[size]=5 will return pagination links like https://api.osf.io/v2/nodes/?page=2&page%5Bsize%5D=5

abought: Hrm. It feels like this code is relying on somewhere other than itself to not override the query by accident, rather than using something like `Object.assign` to merge with sensible defaults.

I'm ok with this merging based on the above- it was the behavior before- but for future reference this seems like a habit we want to avoid. This isn't new to this PR, but it happened to show up in the diff.

PR #166, opened 2017-01-12
[EOSF-458] Feature/yarn

binoculars:
## Ticket
https://openscience.atlassian.net/browse/EOSF-458

# Purpose
Using yarn as the package manager, instead of npm

# Notes for Reviewers
run `brew update && brew install yarn` locally. From now on, you'll use yarn to install dependencies


.template-lintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     extends: 'recommended',
 
     rules: {
-        'block-indentation': 4
+        'block-indentation': 4,
+        'bare-strings': false
```
abought: In theory, we want ember-osf to have full translation support, to best support consuming apps (like preprints) that also have this requirement.

Is there a reason why [bare-strings](https://github.com/rwjblue/ember-template-lint#bare-strings) is disabled here? Is this a permanent or temporary change?

(Temporary is fine. The deprecation warnings are annoying)

binoculars: Temporary--the console warnings dictated that it at least be there, but since if it were set to true, it would cause test failures (I tried it and there were *a lot*). I will add a ticket to the backlog to fix this. I agree with you, we should have translation support.

binoculars: [Jira Ticket](https://openscience.atlassian.net/browse/EOSF-456)

PR #167, opened 2017-01-18
Add fixString helper that makes bad server data look human readable [EOSF-452]

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-452

⚠️ This PR had multiple authors and scope grew. It would be a good candidate for squashing on merge.

# Purpose
Global fix for every string field that comes back from the API with mangled ampersands. (and in the future, maybe other special characters)

This is similar to the global filter `temp_ampersand_fixer` in mako, and centralizes what would otherwise be one-off fixes in every ember template

Deliberately exempt filePath and URL fields from use of this helper. If we decide that those fields also need help, we can apply the filter to those fields later.

# Notes for Reviewers
- Adds methods for dealing with malformed text:
  - `fixSpecialChars` utility function, which powers...
  - ...a template helper (for compound fields like `node.tags`)
  - ...a field transform (which will automatically fix most characters on display, if used in models)
- Fixes a legacy bug in which an API [list field](https://github.com/abought/osf.io/blob/e17666d666beb60dbabfb4eaf652efe523300460/api/nodes/serializers.py#L310-L310) was being serialized as a string. It is expected that this change will be compatible with usages in ember-preprints and this dummy app.
- Minor fixes to resolve YUIDoc doc build errors. Updating the `docs` folder should happen in a separate PR.

# Testing notes 
Setup: Create an OSF node with special characters in the title or description.
- [x] Verify that with no other cleanup, an ember page displaying that node (eg preprints) will display the title: as `Node & stuff` rather than `Node &amp; stuff`. Eg, this PR by itself will fix display of a bad title.
- [x] From the front end, change a field that does not have special characters. Confirm that the original server-provided value is what gets sent back to the server for the fields you did not change. (if you edit description, title sent to server would still be `Node &amp; stuff`- even it it displays differently on the front-end.
- [x] From the front end, change the value of a field that does contain special characters. Confirm that what you wrote is what gets sent to the server.
- [x] Reload the page. Confirm that the new value is displayed on the page as you intended.

For local testing in, eg, ember-preprints, change this line in package.json to the head commit of this branch (below may be out of date):
>    "ember-osf": "git+https://github.com/abought/ember-osf.git#dc787dcf57b258cbe02116ba2c33e5f36dca39ca",

Then `docker-compose stop preprints` and `docker-compose up preprints` to restart the container and pull the new dependency from github.


abought:
Will reassess failing tests after https://github.com/CenterForOpenScience/ember-osf/pull/166 is merged.

PR #169, opened 2017-01-22
Feature/Discover Page Component [EOSF-521]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-521

❗️ Blocking:
https://github.com/pattisdr/retraction-watch
https://github.com/CenterForOpenScience/ember-preprints/pull/269
https://github.com/CenterForOpenScience/ember-preprints/pull/302
https://github.com/CenterForOpenScience/ember-osf-registries/pull/11

# Purpose
This creates a reusable Ember-OSF component that can build the majority of the Discover Page - which is a search interface on top of SHARE.  This component was originally built for Retraction Watch in Jan '17, and then later modified to accommodate Preprints and Registries Discover Pages in Mar '17.  Ember-SHARE search facets are available (and being used in RW), but I don't recommend using them alongside Preprints/Registries search facets until more of the code is combined. The majority of the code was pulled directly from Ember-SHARE and Preprints, with a small amount adapted from Registries.  Discover page has generic analytics for Preprints/Registries items.

# Sample Usage
Many pieces of the page can be customized.  Pass in facets to determine which search facets you want available in your consuming app.  Facets can be either components in your consuming application (like the provider and subject filters in preprints), or can be search facets from Ember-share, like the language filter.   Pass in additional properties like sortOptions (how your results can be sorted),  lockedParams (which query params cannot be changed, like preprints), or override specific text on the page like the searchPlaceholder.  Each individual query param must be passed into the component as a property so it can be reflected in the URL in your consuming application.  Override styling such as colors in your app's stylesheet.

```
{{discover-page
    activeFilters=activeFilters
    clearFiltersButton=clearFiltersButton
    consumingService=consumingService
    detailRoute=detailRoute
    discoverHeader=discoverHeader
    facets=facets
    fetchedProviders=model
    filterMap=filterMap
    filterReplace=filterReplace
    lockedParams=lockedParams
    page=page
    provider=provider
    q=q
    queryParams=queryParams
    searchButton=searchButton
    searchPlaceholder=searchPlaceholder
    showActiveFilters=showActiveFilters
    sortOptions=sortOptions
    subject=subject
}}

```
# Notes for Reviewers
Is demonstrated in EOSF-557 (Registries) and EOSF-558 (Preprints) to build the discover pages.  Is also demonstrated in Retraction Watch (currently housed here https://github.com/pattisdr/retraction-watch).

The discover-page component itself has some duplicate code handling the same things, but different ways.  Ember-SHARE and Ember-Preprints discover pages handled things differently under the hood. Rather than entirely reconcile everything, the goal was to get this component to be able to build the discover page, with all the different facet types available.   For example, Ember-SHARE does AND searches and preprints/registries does OR searches. More work will be done to consolidate when the search pages when there are decisions on how Preprints/Registries search will be improved.

# For QA:
Can be tested in https://openscience.atlassian.net/browse/EOSF-557, https://openscience.atlassian.net/browse/EOSF-558.

Re-test all discover-page functionality.  Behavior should be similar. There are slight UI tweaks now that the same component is building both pages. Testing should include creating network failures to observe what happens when a request to SHARE/API v2 fails.


pattisdr:
hmm. tests passing locally.

abought:
> hmm. tests passing locally.

I've been seeing this error recently as well- it's nondeterministic and often retrying fixes it.

I'd been ignoring it at first as there was some turbulence with yarn, but now it's starting to seem like something worth a deeper look.

pattisdr:
tests are passing locally.

addon/components/discover-page/template.hbs
```diff
@@ -0,0 +1,180 @@
...
+                {{#if showActiveFilters}}
+                    <div id="activeFilters" class="well well-sm p-md">
+                        <div class="row">
+                            <div class="col-sm-6">
+                                <h4>{{t "eosf.components.discoverPage.activeFilters.heading"}}:</h4>
+                            </div>
+                            <div class="col-sm-6">
+                                <span class="pull-right">
+                                    {{!CLEAR FILTERS BUTTON}}
+                                    <button type="button" class="btn btn-default btn-sm" {{action "clearFilters"}}>{{clearFiltersButton}} {{fa-icon "remove"}}</button>
```
abought: Please add aria label for this icon.

There are a number of additional uses of fa-icon throughout this code that will need similar tweaks, but I'll avoid spamming this PR to note them; I know you've always been good at tracking down these small details.

tests/dummy/app/index.html
```diff
@@ -14,6 +14,16 @@
     <link rel="stylesheet" href="assets/vendor.css">
     <link rel="stylesheet" href="assets/dummy.css">
 
+    <script type="text/javascript"
+        src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
+    </script>
+    <script type="text/x-mathjax-config">
```
pattisdr: This is not a dependency for consuming apps - not all apps are going to need mathjax.  For most of my work on this component, i was using the dummy app to test that things were working; in this case, I wanted to see mathjax in the search results.

abought:
> This is not a dependency for consuming apps - not all apps are going to need mathjax.

Would it be fair to say that any apps that use the search widgets will need mathjax?

Even if we don't add this to bower, people building on this addon will need to know something about the requirement in order to use these specific components effectively.

pattisdr:
Sure, your search results will look better if you have mathjax.

tests/dummy/app/index.html
pattisdr: @abought - do you have a preference for where the mathjax instructions should be documented?

abought: That's an excellent question. There are two levels on which we can answer. Especially wrt bower, the likely solution now is a mix of these two answers.

- This addon has never been great to setup from scratch, so the install docs don't have much precedent. Even a README section to start would help.
- For a more realistic solution: the addon should handle the installation of external dependencies for the user. (rather than making them read prose and paste in code) This means giving them the option, and making all the steps clear and traceable:
  - Try to install mathjax via bower if possible. Hard-coding the cdn url means that mathjax is hidden from all the tools we would normally use to find and update our dependencies. If someone finds a security vulnerability- we would not know it affected us.
    - Addons are able to tell a consuming app that they (may) need a bower dependency, [during the addon install process](https://ember-cli.com/api/classes/Blueprint.html#method_addBowerPackageToProject) (using hooks in the default blueprint). The consuming app author could then decide whether to accept or remove that bower.json entry. 
  - Some addons provide configuration options that can control whether certain features are used. For the `script` block, ideally the addon would provide the configuration it needs when the app loads. Not immediately sure of the best mechanism for doing this (maybe an initializer of some kind), but see, eg [ember-bootstrap docs](http://www.ember-bootstrap.com/#/getting-started/assets) for inspiration on how this might work.

pattisdr: @abought maybe there was a reason MathJax was being installed this way? https://github.com/CenterForOpenScience/ember-preprints/pull/129/files#diff-b9707fe88797285e93483456d1389b61

abought: Sometimes we'll use a CDN in prod but a packaged dependency locally. (to make debugging easier) Eg:
https://github.com/CenterForOpenScience/ember-preprints/blob/develop/ember-cli-build.js#L62

The biggest change in my thinking since the last review has been our interest in automated tools for dependency management. Things like Snyk (security scanner) and Greenkeeper (package.json updater) rely on these explicit file listings of dependencies to figure out what your project uses, and that makes hard-coded URLs less attractive. 

We're increasingly noticing the challenge of finding and incorporating updates as we do this more. To give you a concrete example? The [latest version](http://docs.mathjax.org/en/latest/whats-new-2.7.html) of MathJax is actually 2.7. (Oct 2016)

pattisdr: I see.  So the new task is probably trying to getting bower to install MathJax.  I think this should be a follow-up Jira ticket.  Preprints and Registries are currently using CDN.

tests/dummy/app/templates/share-discover-page.hbs
```diff
@@ -0,0 +1,33 @@
+{{title 'Search'}}
+
+{{discover-page
+    clearFiltersButton="Clear filters"
```
pattisdr: I was providing more options for customization, but have since removed a few of them.  This is not so many configuration options given that the component builds an entire page. Especially since half of the properties passed in are the individual query parameters so when they are updated in the component, they will be reflected in the URL.

tests/dummy/config/environment.js
```diff
@@ -42,6 +42,12 @@ module.exports = function(environment) {
         baseURL: '/',
         locationType: 'auto',
 
+        SHARE: {
```
pattisdr: Thanks for pointing out this file - I wasn't aware of it.

tests/index.html
```diff
@@ -13,8 +13,16 @@
     <link rel="stylesheet" href="assets/vendor.css">
     <link rel="stylesheet" href="assets/dummy.css">
     <link rel="stylesheet" href="assets/test-support.css">
-
     {{content-for "head-footer"}}
+    <script type="text/javascript"
```
pattisdr: As above, I think installing MathJax via bower should be a separate ticket.  I experimented for a little while and it looked like it was going to take some work.  Preprints and Registries are currently using CDN - this copied bit was just so I could get the dummy app to look like preprints.

package.json
```diff
@@ -54,7 +54,6 @@
     "ember-collection": "1.0.0-alpha.5",
     "ember-data": "2.7.0",
     "ember-data-factory-guy": "^2.6.1",
-    "ember-disable-prototype-extensions": "^1.0.0",
```
pattisdr: Ember-SHARE was using prototype extensions and the majority of this PR was directly pulled from that repo. Prototype extensions worked fine in preprints and registries, so I removed this addon.  

However, I have rewritten the sections of this PR that are using prototype extensions and restored the addon.

addon/components/discover-page/template.hbs
pattisdr: This icon is on the "Clear Filters" button, which says "Clear Filters".  The remove icon is aria-hidden=true.   I don't think the user needs the remove icon to be visible since the button itself takes care of it.  I have added other aria-labels where it makes sense for other font-awesome icons - thanks for the reminder.

addon/components/discover-page/template.hbs
```diff
@@ -0,0 +1,180 @@
...
+                        </div>
+                    </div>
+                {{/if}}
+
+                {{!SEARCH FACETS}}
+                {{#if showActiveFilters}}
+                    <h4> {{t 'eosf.components.discoverPage.refineSearch'}} </h4>
+                {{/if}}
+                {{faceted-search
+                    onChange='filtersChanged'
```
pattisdr: If the network request is a 400, we show Lucene Search Help.  If a network request fails for any other reason, they get a message saying that Search is Unavailable.  I don't know if I agree that the search-facet components should be reset.  Several times as I was developing this, Staging SHARE would be down for a few seconds and my search would fail. I could just retry the search because my filters were still there. Regardless, Preprints and Registries are not using these actions right now, these are for the Ember-SHARE facets.

addon/components/search-facet-worktype/component.js
```diff
@@ -0,0 +1,61 @@
...
+
+    selected: Ember.computed('state', function() {
+        return this.get('state') || [];
+    }),
+
+    changed: Ember.observer('state', function() {
+        let state = Ember.isBlank(this.get('state')) ? [] : this.get('state');
+        let previousState = this.get('previousState') || [];
+
+        if (Ember.compare(previousState, state) !== 0) {
```
pattisdr: I just copied this from Ember-SHARE, but I am guessing it was used to be able to compare arrays.  Even if both arrays are equal, comparing previousState !== state will be true.

package.json
abought: Thanks.

I recently had addon unit tests fail completely because the addon code was using prototype extensions. (and ember unit tests turn those extensions off) 

So as I added tests, more stuff was breaking- it was certainly no fun to track down, and I'd love to save someone else that trouble. Appreciate your help and improvements to the code.

addon/components/search-result/style.scss
```diff
@@ -0,0 +1,53 @@
+.search-result {
+    background: white;
+    .row {
+        font-family: 'Open Sans' , 'Helvetica Neue' , sans-serif;
+        -webkit-font-smoothing: antialiased;
```
abought: See note elsewhere about avoiding `font-smooth` usage if possible.

addon/components/search-facet-locked/style.scss
```diff
@@ -0,0 +1,3 @@
+.locked-items {
```
abought: This style appears to be blank. Can it be removed?

addon/components/search-facet-worktype-button/template.hbs
```diff
@@ -0,0 +1,12 @@
+{{!Copied from Ember-SHARE}}
+<div class='type-filter-option {{if selected 'active'}}'>
+    {{#if collapsible}}
+        <i class='fa expand-icon clickable {{if toggleState 'fa-plus' 'fa-minus'}}' {{action 'toggleBody'}}></i>
+    {{/if}}
+    <span class={{if selected 'bubble'}}>
+        <a class='clickable' {{action 'click'}}>{{label}}</a>
```
abought: Consider changing all interactive elements in this file to `buttons` instead of `a` or `i` elements.

As elsewhere, I would also recommend adding `aria-labels` to at least some of the elements in this file, so they are not invisible to screen readers.

addon/components/search-help-modal/component.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
+    layout,
+    isOpen: false,
+    currentPath: Ember.computed(function() {
+        return Ember.getOwner(this).lookup('controller:application').currentPath;
```
abought: Does this imply that this component requires additional application configuration to use? (special properties on the application controller)

In general we try to keep components reusable by explicitly passing them the info they need, rather than relying on the code of a controller on a different route entirely. I'd recommend refactoring here if possible.

addon/components/search-help-modal/component.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
```
abought: In general, we try to add basic docstrings to all new components (what they are, what they do, what params they need). I tend to encourage this much more strongly in ember-osf, simply because we intend this repo to be highly reusable shared code.

config/backends.json
```diff
@@ -15,7 +18,10 @@
         "waterbutlerUrl": "https://staging-files.osf.io/",
         "helpUrl": "http://help.osf.io",
         "cookieLoginUrl": "https://staging-accounts.osf.io/login",
-        "oauthUrl": "https://staging-accounts.osf.io/oauth2/authorize"
+        "oauthUrl": "https://staging-accounts.osf.io/oauth2/authorize",
```
abought: Please also add corresponding property names to the `env` section (line 65) so that these new URLs can optionally be overridden as environment variables.

tests/integration/components/search-facet-locked/component-test.js
```diff
@@ -0,0 +1,16 @@
...
+});
+
+test('it renders', function(assert) {
+
+  // Set any properties with this.set('myProperty', 'value');
+  // Handle any actions with this.on('myAction', function(val) { ... });
+
+  this.render(hbs`{{search-facet-locked}}`);
+
+  assert.equal(this.$().text().trim(), '');
```
abought: Is there perhaps a more meaningful test we could do here?

tests/unit/helpers/sort-option-display-test.js
```diff
@@ -0,0 +1,12 @@
+
+import { sortOptionDisplay } from 'dummy/helpers/sort-option-display';
+import { module, test } from 'qunit';
+
+module('Unit | Helper | sort option display');
+
+test('returns corresponding display text', function(assert) {
```
abought: I just wanted to say thanks again for your work adding tests to this code- both the new and inherited parts. I know there will be a ways to go yet (this is a very complex feature!).... but this will make a big step forward. Your diligence and care is really appreciated.

addon/services/theme.js
```diff
@@ -0,0 +1,30 @@
+import Ember from 'ember';
```
abought: Note for line above: do we really need to add this logo twice to the same repo?

`share-logo-icon.png`

tests/unit/helpers/number-format-test.js
```diff
@@ -0,0 +1,17 @@
+import { numberFormat } from 'dummy/helpers/number-format';
+import { module, test, skip } from 'qunit';
+
+module('Unit | Helper | number format');
+
+// Works in Chrome but doesn't work in terminal
```
pattisdr: This is a test of a helper that just calls number.toLocaleString(), so that is all we are testing here.  It looks like phantomJS does not apply the locale formatting correctly. https://github.com/ariya/phantomjs/issues/12581. I think it works fine across browsers.  I modified the test per someone's suggestion on that issue.  Skipping the test if it's running in PhantomJS but running elsewhere.

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+        // Ember-SHARE property. Returns pages of hidden search results.
+        const total = this.get('totalPages');
+        const max = this.get('clampedPages');
+        if (total !== max) {
+            return total - max;
+        }
+        return null;
+    }),
+    providerChanged: Ember.on('init', Ember.observer('provider', function() {
+        // For PREPRINTS and REGISTRIES - watches provider query param for changes and modifies activeFilters
```
pattisdr: The observer is something brought over from preprints.  I experimented with some other ways of doing this, but right now this is one of the main ways the application is notified that something has changed and search needs to be reloaded.

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+                loading: false,
+                firstLoad: false,
+                results: results,
+                queryError: false,
+                shareDown: false,
+            });
+            if (this.get('totalPages') && this.get('totalPages') < this.get('page')) {
+                this.search();
+            }
+        }, (errorResponse) => {
```
pattisdr: I didn't think you could do .then().catch() after jQuery AJAX requests.  I changed to .then().fail()

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+        if (this.get('theme.isProvider')) {
+            this.get('theme.provider').then(provider => {
+                this.set('providerName', provider.get('name'));
+                this.loadPage();
+            });
+        }
+    },
+    scrollToResults() {
+        // Scrolls to top of search results
+        Ember.$('html, body').scrollTop(Ember.$('.results-top').position().top);
```
pattisdr: this.$() worked on the second usage.

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+                    label: 'Discover - Search',
+                    extra: this.get('q')
+                });
+
+            this.search();
+        },
+        selectSortOption(option) {
+            // Runs when sort option changed in dropdown
+            this.set('sort', option);
+            Ember.get(this, 'metrics')
```
pattisdr: I just brought this over from preprints but my guess is that it was using same conventions from the ember-metrics docs https://github.com/poteto/ember-metrics#usage

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+                return;
+            }
+            // Tracks search on keypress, debounced
+            Ember.run.debounce(this, this.trackDebouncedSearch, 3000);
+            this.search();
+        },
+        updateFilters(filterType, item) {
+            // For PREPRINTS and REGISTRIES.  Modifies activeFilters.
+            item = typeof item === 'object' ? item.text : item;
+            const filters = Ember.$.extend(true, Ember.A(), this.get(`activeFilters.${filterType}`));
```
pattisdr: Even though this deep copies, I also need to put it into an Ember Array here so I can do push/removeObject shortly after.

addon/components/faceted-search/style.scss
```diff
@@ -0,0 +1,56 @@
+/* Override in consuming application to customize appearance of search facets */
+
+.panel-default {
+    font-family: 'Open Sans' , 'Helvetica Neue' , sans-serif;
+    -webkit-font-smoothing: antialiased ;
```
pattisdr: OK - removed. I copied this style from Ember-SHARE and assume it was done intentionally.  I can't really tell the difference in the smoothness with or without it so I will remove it. I might be missing something.

addon/components/search-help-modal/component.js
pattisdr: Nothing needs to be added to the consuming app's Application Controller.  currentPath is already there https://guides.emberjs.com/v1.10.0/understanding-ember/debugging/#toc_get-current-route-name-path.  Different pages where this search modal will be used have different path names.  The examples in the search modal link build a search query for you. I need to know the currentPath. I can also pass in currentPath:"discover" or similar, alternatively.

addon/services/theme.js
pattisdr: good catch.

tests/unit/helpers/number-format-test.js
abought: Per the issue in the linked ticket: according to the MDN docs the function parameters were added later, and compatibility with that would be around the edge of our supported browser range. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#Browser_compatibility

But you don't appear to be using anything more than the stock older variant; I'm a little curious what the failure message looks like. So long as it's purely PhantomJS being kooky, then this is ok to skip for now- I'll make a note to take a look during the next review.

addon/components/discover-page/component.js
abought: That's a good point- deferreds are slightly different, and .fail would probably work. The main goal is to catch errors that happen inside the `.then` block.

To help us treat all requests via the same promise semantics, I typically do something like the following; we really should abstract this to a helper function at some point. (especially as we add more code to make manual AJAX requests- I'd much prefer one consistent API for interacting with all promises!)

```
        let jqDeferred = this._doRequest();
        return new Ember.RSVP.Promise((resolve, reject) => {
            jqDeferred.done((value) => resolve(value));
            jqDeferred.fail((reason) => reject(reason));
        });
```

Addons like `ember-ajax` do something [similar](https://github.com/ember-cli/ember-ajax/blob/master/addon/mixins/ajax-request.js#L212-L218).

tests/unit/helpers/number-format-test.js
pattisdr: @abought, the linked ticket it is showing the exact thing that i'm seeing.  If I do toLocaleString(1234) in phantomJS, it just returns 1234.  In the browser, it would be 1,234.

addon/components/faceted-search/style.scss
```diff
@@ -0,0 +1,55 @@
...
+    opacity: 1;
+}
+
+::-moz-placeholder { /* Mozilla Firefox 19+ */
+    color: #333333;
+    opacity: 1;
+}
+
+:-ms-input-placeholder { /* Internet Explorer 10-11 */
+    color: #333333 ;
```
abought: Thanks for catching browser differences... but are we missing the non-prefixed version of this pseudo-element? (aka the standardized one we secretly wanted all along)

`::placeholder`

For future reference, and solely to share something that can help with a common CSS headache, see:
https://github.com/postcss/autoprefixer

I think we experimented with postcss in preprints before. It might simplify life going forward.

addon/components/query-syntax/template.hbs
```diff
@@ -0,0 +1,101 @@
+{{!Copied from Ember SHARE}}
+<div class="query-syntax-help">
+    <h2>Could not perform search query.</h2>
```
abought: Hard-coded text. We may want to consider moving it to a translation file.

addon/components/search-facet-typeahead/component.js
```diff
@@ -0,0 +1,125 @@
...
+
+        return Ember.$.ajax({
+            url: this.typeaheadQueryUrl(),
+            crossDomain: true,
+            type: 'POST',
+            contentType: 'application/json',
+            data: data
+        }).then((json) =>
+            resolve(this.handleTypeaheadResponse(json)),
+            reject
```
abought: What is the purpose of the `,reject` here? 

Additionally, passing resolve and reject handlers into the call signature of a function that does its own AJAX work seems a bit convoluted. This could perhaps be simplified for maintainability.

addon/components/search-facet-worktype-button/template.hbs
```diff
@@ -0,0 +1,16 @@
...
+    {{#if collapsible}}
+        <button class="btn-link" aria-label={{t 'eosf.components.searchFacetWorktypeButton.expandWorktype'}}>
+            <i class='fa expand-icon clickable {{if toggleState 'fa-plus' 'fa-minus'}}' {{action 'toggleBody'}}></i>
+        </button>
+    {{/if}}
+    <span class={{if selected 'bubble'}}>
+        <button class='clickable btn-link' {{action 'click'}}>{{label}}</button>
+        {{#if selected}}
+            <button class="btn-link" aria-label={{t 'eosf.components.searchFacetWorktypeButton.removeWorktypeFilter'}}>
+                <i class='fa fa-times close-icon clickable' {{action 'click'}}></i>
```
abought: Click action handler might go better on the `button` than the `i`con.

That said, thank you for setting me up for that HTML syntax pun.

addon/components/search-result/component.js
```diff
@@ -0,0 +1,174 @@
...
+    }),
+
+    type: Ember.computed('result.type', function() {
+        return this.get('result.type').replace(/\w\S*/g, function(str) {return str.capitalize();});
+    }),
+    safeTitle: Ember.computed('result.title', function() {
+        return Ember.String.htmlSafe(this.get('result.title')).string;
+    }),
+    safeDescription: Ember.computed('result.description', function() {
+        return Ember.String.htmlSafe(this.get('result.description')).string;
```
abought: I very strongly recommend against using this method. Perhaps a better name should be `Ember.String.weReallyReallyTrustTheUserEnteredOkTextOnThisOneOtherwiseWhoopsie`.

The only time this should *ever* be used is for text from a trusted source that we completely control (like a string of advisory board members from the API). I suspect the description comes from some external input and therefore should not be marked as safe. This is especially true for SHARE, where the text we harvest comes from *many* external sources that all have different rules.

Can you clarify where this came in? I suspect the appropriate action will be a combination of the following:

1. Remove all uses of `htmlSafe` on user-entered text within this PR,
2. Send a note to whichever upstream repo this was borrowed from and make sure the fix gets propagated.
3. Identify the reason this was ever used, and develop alternatives to fit the intended use case. For example, most of our ember-data models clean up bad ampersands via a custom [field type](https://github.com/abought/ember-osf/blob/6627dbefd88fbefdc176a3ae300c45347b2cfd5c/addon/transforms/fixstring.js#L28-L28), also available as a [template helper](https://github.com/abought/ember-osf/blob/6627dbefd88fbefdc176a3ae300c45347b2cfd5c/addon/helpers/fix-special-char.js#L16-L16) for your convenience

addon/components/search-result/component.js
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            return 0;
+        }
+    },
+    actions: {
+        addFilter(type, filter) {
+            this.sendAction('addFilter', type, filter);
+        },
+        toggleShowBody() {
+            this.set('showBody', !this.showBody);
```
abought: This can be simplified as `this.toggleProperty('showBody')`

For consistency, we generally use `this.get()` for all properties we fetch. (vs `this.propertyName`) Booleans are a primitive property, but if we ever changed this field to a computed, using `this.get` makes it easier to refactor.

addon/helpers/sort-option-display.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+/**
```
abought: Verify that this file shows up correctly in built docs. It may need the `@module` annotation.

addon/utils/elastic-query.js
```diff
@@ -0,0 +1,93 @@
+import Ember from 'ember';
+import moment from 'moment';
+
+/*
```
abought: Please verify that this shows up correctly in built docs.

tests/dummy/app/controllers/share-discover-page.js
```diff
@@ -0,0 +1,44 @@
...
+    end: '',
+    i18n: Ember.inject.service(),
+    institutions: '',
+    filterMap: {
+        providers: 'sources',
+        subjects: 'subjects'
+    },
+    filterReplace: {
+        'Open Science Framework': 'OSF',
+        'Cognitive Sciences ePrint Archive': 'Cogprints',
```
abought: Are there any translation concerns for these strings?

Not urgent, but noting while it's in front of us.

addon/components/discover-page/template.hbs
abought: More generally, I will encourage adding "create network failures" to the QA testing notes when this goes in.

addon/components/search-result/component.js
pattisdr: @abought - this was from Ember-SHARE.  Will address the rest of this comment later.

README.md
```diff
@@ -134,6 +134,22 @@ to fetch all nodes (or at least the first page of results). If you need to fetch
 information about how to handle pagination. Ember-osf also provides support for 
 [paginated relationship requests](https://github.com/mdehoog/ember-data-has-many-query) via a third-party addon.
 
+### MathJax
+
+We use [MathJax](https://www.mathjax.org/) to make math look pretty in all browsers.  If you want this in your application, copy this section into the `<head>` element of your `index.html` file:
+
+```
+<script type="text/javascript"
+    src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
```
pattisdr: Follow-up ticket here: https://openscience.atlassian.net/browse/EOSF-602

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+     */
+    discoverHeader: null,
+    /**
+     * End query parameter.  If "end" is one of your query params, it must be passed to the component so it can be reflected in the URL.
+     * @property {String} end
+     */
+    end: '',
+    /**
+     * A list of the components to be used for the search facets.
+     * @property {Array} facets
```
pattisdr: Facets is something the user passes in, which is why I added it to the YUIDOC comments.  I made the default a computed property, which is how Ember-SHARE is doing this, but user can override. This is how preprints and registries can specify which search facets are available.

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+            filters.push({
+                terms: {
+                    sources: [this.get('providerName')]
+                }
+            });
+        }
+
+        let query = {
+            query_string: {
+                query: this.get('q') || '*'
```
pattisdr: This is how Ember-SHARE and preprints are currently writing this. I think this is so a blank text field can query.

addon/components/discover-page/component.js
```diff
@@ -0,0 +1,808 @@
...
+            let restoreActiveFilters = {};
+            Object.keys(this.get('activeFilters')).forEach(filter => {
+                if (filter === 'providers') {
+                    restoreActiveFilters[filter] = this.get('theme.isProvider') ? this.get('activeFilters.providers') : [];
+                } else {
+                    restoreActiveFilters[filter] = [];
+                }
+
+            });
+            this.set('activeFilters', restoreActiveFilters);Ember.get(this, 'metrics')
```
pattisdr: oops, thanks.

addon/locales/en-us/translations.js
```diff
@@ -18,6 +18,59 @@ export default {
...
+            discoverPage: {
+                activeFilters: {
+                    heading: 'Active Filters',
+                    button: 'Clear filters'
+                },
+                search: 'Search',
+                searchPlaceholder: 'Search...',
+                poweredBy: 'powered by',
+                noResults: 'No results. Try removing some filters.',
+                shareTotalText: `results`,
```
pattisdr: Actually removing this - no longer being used.

addon/components/faceted-search/style.scss
pattisdr: I just copied this from Ember-Share.  I am not sure what you mean when you say 

> "missing the non-prefixed version of this pseudo-element"

but it doesn't look like you can put ::placeholder in the stylesheet.  It looks like you style placeholder text with the collection of vendor-prefixed selectors that are in this stylesheet.

addon/components/search-facet-typeahead/component.js
pattisdr: Did not write this - copied from Ember-SHARE. However, it looks like passing in the reject handler seems to allow the component to update if the request fails.  For example, there is a "People" search facet.  If SHARE is down, and I attempt to search, the "reject" allows me to display "No results found". 
screen shot 2017-04-10 at 12 28 03 pm

 If there is no "reject", then the component does not update if the search fails. It continues to say "Loading options".
screen shot 2017-04-10 at 12 30 07 pm

addon/components/search-result/component.js
pattisdr: Will be removed.  Related ticket here https://openscience.atlassian.net/browse/SHARE-755

tests/dummy/app/controllers/share-discover-page.js
pattisdr: Do we translate proper nouns? Seems to me like Open Science Framework should be Open Science Framework everywhere. It's the name of our product.

addon/components/search-result/component.js
```diff
@@ -0,0 +1,170 @@
...
+     * Array of query params being used in consuming app
+     * @property {Array} queryParams
+     */
+    queryParams: null,
+    /**
+     * Search result from SHARE
+     * @property {Object} result
+     */
+    result: null,
+    providerUrlRegex: {
```
abought: With the advent of branding etc, I would encourage moving provider info to a single standard location. Having to maintain the list of providers in many different components could quickly pose a challenge for maintainability.

addon/components/search-result/component.js
```diff
@@ -0,0 +1,170 @@
...
+        let re = null;
+        for (let i = 0; i < this.get('result.providers.length'); i++) {
+            //If the result has multiple providers, and one of them matches, use the first one found.
+            re = this.providerUrlRegex[this.get('result.providers')[i].name];
+            if (re) break;
+        }
+
+        re = re || this.providerUrlRegex.OSF;
+
+        const identifiers = this.get('result.identifiers').filter(ident => ident.startsWith('http://'));
```
abought: Will identifiers ever start with `https`?

(https by default is getting increasingly more common)

addon/components/search-result/template.hbs
```diff
@@ -0,0 +1,124 @@
...
+        <div class="col-xs-12">
+            {{!Title}}
+            <h4>
+                {{#if (and detailRoute osfID)}}
+                    {{#link-to detailRoute osfID invokeAction=(action 'click' 'link' 'Discover - Result Title' osfID)}}{{fix-special-char result.title}}{{/link-to}}
+                {{else if (and detailRoute (eq hostAppName 'Retraction Watch'))}}
+                    {{#link-to detailRoute result.id}}{{fix-special-char result.title}}{{/link-to}}
+                {{else}}
+                    {{#if hyperlink}}
+                        <a href='{{hyperlink}}' onclick={{action 'click' 'link' (concat 'Discover - Result Title: ' hyperlink) result.id}}>{{fix-special-char result.title}}</a>
```
abought: Possible translation needed (various points within this template).

addon/components/search-result/template.hbs
```diff
@@ -0,0 +1,124 @@
...
+                        {{#each result.infoLinks as |link|}}
+                            <li><b>{{link.type}}:</b> {{link.uri}}</li>
+                        {{/each}}
+                    </ul>
+                {{/if}}
+
+                {{!Tags}}
+                {{#each tags as |tag|}}
+                    {{#if tagsInQueryParams}}
+                        <a class="filter" {{action 'addFilter' 'tags' tag}}>
```
abought: Anchor should be button.

tests/dummy/app/controllers/share-discover-page.js
abought: That's a good question. Might depend on the product. Eg I could imagine translating "research papers in economics" while leaving the short version as `RePEc`.

addon/components/query-syntax/template.hbs
```diff
@@ -0,0 +1,100 @@
...
+            <li><code>girafe~2</code> -- {{t 'eosf.components.querySyntax.match'}} "giraffe", "gyrate", "pirate", etc.</li>
+        </ul>
+    </div>
+    <div>
+        <h3>{{t 'eosf.components.querySyntax.phraseProximity'}}</h3>
+        <p>
+            {{t 'eosf.components.querySyntax.phraseProximityDesc1'}}
+        </p>
+        <ul>
+            <li><code>"giraffe neck"~2</code> -- {{t 'eosf.components.querySyntax.match'}} "giraffe has a neck", {{t 'eosf.components.querySyntax.butNot'}} "giraffe has a long neck".</li>
```
pattisdr: I chose not to translate these examples because I was concerned about them being translated properly.

addon/components/query-syntax/template.hbs
```diff
@@ -0,0 +1,100 @@
...
+            {{t 'eosf.components.querySyntax.boostingDesc1'}} <code>^</code> {{t 'eosf.components.querySyntax.boostingDesc2'}}
+        </p>
+        <ul>
+            <li><code>giraffe^2 lion</code> -- {{t 'eosf.components.querySyntax.boostingDesc3'}} "giraffe" {{t 'eosf.components.querySyntax.booleanDesc4'}} "lion".</li>
+        </ul>
+    </div>
+    <div>
+        <h3>{{t 'eosf.components.querySyntax.moreInformation'}}</h3>
+        <p>
+            {{t 'eosf.components.querySyntax.moreInfoQuerySyntax'}} <a href='https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax'>Elasticsearch documentation</a>.
```
pattisdr: I can translate documentation, but I don't think we should translate Elasticsearch.

addon/components/search-result/template.hbs
pattisdr: `Discover - Result Title:` is being sent to Google Analytics and does not need translation.

tests/dummy/app/controllers/share-discover-page.js
pattisdr: I do not believe we should translate these, but I will put up follow-up ticket for this and see what others think.

addon/components/discover-page/template.hbs
```diff
@@ -0,0 +1,182 @@
...
+                {{#if (eq hostAppName 'Preprints')}}
+                    {{add-preprint-box}}
+                {{/if}}
+
+            </div>
+
+            <div class="col-sm-8 col-xs-12">  {{!RIGHT PANEL - SEARCH RESULTS}}
+                {{#if loading}} {{!SEARCH LOADING}}
+                    <div class="text-center p-v-md" aria-label={{t 'eosf.components.discoverPage.searchLoading'}}>
+                        {{fa-icon 'spinner' ariaHidden=false pulse=true size=3}}
```
pattisdr: Ticketed here https://openscience.atlassian.net/browse/EOSF-626

tests/dummy/app/controllers/share-discover-page.js
pattisdr: We don't seem to be displaying the long version?

PR #170, opened 2017-01-25
[rfr] Default authorizationType (and document secret config param) [#EOSF-105]

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-105

# Purpose
Ember-osf is designed to work with two types of authorization: cookies and tokens.

Previously, getting a new app to work required manually setting a secret config parameter, `authorizationType`.

This PR allows the addon to set a [default value](https://ember-cli.com/api/classes/Addon.html#method_config) to that consuming apps don't need more boilerplate to use OAuth 2.0. For internal COS products based on cookie auth, we add documentation of the configuration changes needed to get things working.

# Notes for Reviewers
## How to test this
This PR is about the new developer experience- so the best way to test it is using a fresh ember app. You can probably test this by modifying the `ember-osf` dummy app [config file](https://github.com/abought/ember-osf/blob/5ccb027abe248e3d4fa7f8fc15b1d9cac168f3fc/tests/dummy/config/environment.js#L6-L6). For a more rigorous test, I have created a [standalone dummy app](https://github.com/abought/demo-ember-osf/) that may be useful. Either way...

To test this PR:
1. Clone `ember-osf` and check out this branch locally.
2. Either remove the default value from the dummy app config file, or if using a fresh app, make sure that`package.json` points to the code for this PR on your local computer: `"ember-osf": "file:../ember-osf",`
3. Start the server, go to, eg, `localhost:4200/login`, and try to click the login button with no `authorizationType` provided. It should just work.
4. Point the app at an old version of ember-osf. Before this PR, a missing value of `authorizationType` would cause a JS console error and prevent the user from logging in. With this change, login should work.

## See also
- To verify the docs, see: [Example of cookies being used with ember-osf](https://github.com/abought/ember-preprints/blob/1999349631e98a09f368b469b91193e8b2c57fc3/config/environment.js#L4-L4)

## Routes Added/Updated
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,22 @@ If for some reason you don't have a config/local.yml you can generate one. To do
 ember generate ember-osf
 ```
 
+#### Advanced usage: Selecting an authorization type
+We expect that most projects based on `ember-osf` will authenticate via OAuth 2.0 ("Token Login"); the addon is 
+configured to use this out of the box, so long as you provide your own login page based on the appropriate mixins.
+
+If you are developing an application that will be hosted under the `osf.io` domain, you may wish to use cookie-based 
+authentication instead. In that rare case, add the following lines to your `config/environment.js` file:
+
+```javascript
+    ENV.authorizationType = 'cookie';
+    
+    ENV['ember-simple-auth'] = {
+        authorizer: `authorizer:osf-${defaultAuthorizationType}`,
+        authenticator: `authenticator:osf-${defaultAuthorizationType}`
+    };
+```
+
 ## Usage
 
 #### Ember Data: Using the OSF models
--- a/README.md
+++ b/README.md
@@ -86,11 +86,12 @@ If you are developing an application that will be hosted under the `osf.io` doma
 authentication instead. In that rare case, add the following lines to your `config/environment.js` file:
 
 ```javascript
-    ENV.authorizationType = 'cookie';
+    var authorizationType = 'cookie';
+    ENV.authorizationType = authorizationType;
     
     ENV['ember-simple-auth'] = {
-        authorizer: `authorizer:osf-${defaultAuthorizationType}`,
-        authenticator: `authenticator:osf-${defaultAuthorizationType}`
+        authorizer: `authorizer:osf-${authorizationType}`,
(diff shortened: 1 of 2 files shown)


abought:
> How will this work for the Preprints app since it will be hosted under both osf.io and separately, on the provider domains? I would like to keep the same build and set the authorization type at runtime.

Short answer: this doesn't change the current situation, it just reduces boilerplate by providing a default (and the consuming app can still always choose to override it).

Long answer: you ask an excellent question. I think it may be possible, but it depends on the internals of another package. By default ember simple auth assumes [settings are provided at app build](http://ember-simple-auth.com/api/classes/Configuration.html).

Below are some notes that might be helpful for this discussion.

Internally, the setting is used in several places. Each piece of boilerplate is necessary, but I've never really been happy with the resulting accumulation.

1. When the app first loads, the setting is used to decide whether the login button + behaviors should optimize for token vs cookie auth.
  - https://github.com/abought/ember-osf/blob/f7888191bc6e897e6766f69f34366163ab5364ed/addon/mixins/osf-agnostic-auth-controller.js#L14-L14
  - https://github.com/abought/ember-osf/blob/87e66c8ae5b394d2fbd99ff579bbd26435612ea2/addon/utils/auth.js#L54-L54

2. Right in this PR, it sets the config settings for ember simple auth. All our adapters are based on a mixin that dictates what authorizer to use:
  - http://ember-simple-auth.com/api/classes/DataAdapterMixin.html

3. A very select number of network calls in this app talk to servers without going through the adapter, and need to make decisions about how to send requests. Esp waterbutler related.
  - https://github.com/abought/ember-osf/blob/b93467d36e3248a176aeb2b8cfaaf68a93757a18/addon/utils/ajax-helpers.js#L25-L25
  - https://github.com/abought/ember-osf/blob/81106566da915d8fe355423a3335423990725269/addon/components/dropzone-widget/component.js#L52-L52
  - https://github.com/abought/ember-osf/blob/69fb8cf8e510b70a4e72cc5d4dbb7ea8d26cf1d2/addon/services/file-manager.js#L397-L397

PR #171, opened 2017-01-27
Support environment variables for config + backend URLs [#EOSF-449]

abought:
# Ticket
https://openscience.atlassian.net/browse/EOSF-449

# Purpose
This ticket is intended to make deployments easier, by making it possible to update certain internal configuration without needing to release an entire new version of the addon. This ticket adds support for environment variables, and simplifies the `config/local.yml` blueprint. 

It is intended that these changes be accomplished in a backwards compatible way: it should add features, without breaking existing usages of ember-osf by COS staff or third-party partners.

# Notes for Reviewers
You can always override auth-related settings via environment variables, but attempts to override server URLs will be ignored unless you explicitly specify `BACKEND=env` (see "Running" section of README for an example).

## TODO
- [x] Improve config documentation in README
- [x] Full manual testing for backwards-compatible config file behavior

## This change is developer facing, not user-facing
If this works as intended, users will never notice the change. Basic actions like login, fetching info from the OSF, and clicking "help" should just work as they always did.

It will not be possible for QA to test this change as a standalone, because ember-osf is a library of reusable code. There are no staging servers that deploy ember-osf by itself. We can revisit testing strategy if needed.

## Backwards compatibility!
A major goal is backwards compatibility. This app should support three basic configuration scenarios:

### Old style config/local.yml
The old style had nested sections based on various backend names.
See [old blueprint example](https://github.com/CenterForOpenScience/ember-osf/blob/834b0232d6a2dc2a2a42bd4a01141541fe41e6cb/blueprints/ember-osf/files/config/_local.yml).

### New style config/local.yml
The new file format is flat. See [blueprint](https://github.com/abought/ember-osf/blob/cbcb546a0a2edf5e900b1d55dc96cbc8eb3869a9/blueprints/ember-osf/files/config/_local.yml#L1-L0) for an example.

You can generate a new file by running `ember generate ember-osf`. Be sure to save your old config settings first!

This means it will be slightly harder to switch between OAUTH settings for various server backends, but in our evaluation, this is a rather unusual operation.

#### Any individual OAuth parameter can be overridden via env variables
Can specify some in config/local.yml  , and others by passing them in via environment variables.

Overriding these settings via environment variables will only work if you are using a new-style config file. 

#### Overriding URLs is all-or-nothing
To override the URLs, must specify `BACKEND=env` , as well as one environment variable per  (using the environment variables as named in the above config file). If you use any other [value of BACKEND](https://github.com/abought/ember-osf/blob/cbcb546a0a2edf5e900b1d55dc96cbc8eb3869a9/config/backends.json#L1-L1), your URL environment variables will be ignored. This enables [overriding](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables) of settings via env vars of the [same name](https://github.com/abought/ember-osf/blob/cbcb546a0a2edf5e900b1d55dc96cbc8eb3869a9/config/custom-environment-variables.json#L1-L0), and if an existing server url is changed, that would be a big deal- it would merit a release of ember-osf anyway.

## Console error messages for bad config
When running `ember build` or `ember serve` on the command line, the developer should see a warning message on the console in the following situations.
- [x] The user selects an unknown backend name. (see [available keys](https://github.com/abought/ember-osf/blob/feature/eosf449-dotenv/config/backends.json#L1)). 
- [x] The user specifies the `env` backend, but does not provide a value for all the required environment variables.
- [x] Whenever a user selects the "production" backend while running locally.
- [x] Ensure that app can correctly talk to the server with the modified list of default oauth scopes. (read scope should be unnecessary, as write scope implicitly provides read permissions)

## Summary of tests
For both standalone mode and consuming apps using this addon...
  - [x] Standalone
  - [x] Consuming app

Checklist: 
- [x] Specify OAuth settings in old config file format, confirm they are used.
- [x] Specify OAuth settings in new config file format, and confirm they are used.
  - [x] Verify that one (or all!) values can be overridden with an env var of the same name
- [x] Specify known, and unknown, backend names, and confirm expected behaviors.
- [x] Specify `BACKEND=env` + all the required URLs, and confirm that app works
  - [x] Provide them via config file
  - [x] Provide them via env vars
  - [x] A mix of the two
  - [x] If one of those urls is omitted, there should be console errors.
  - [x] Specify `BACKEND=env`, but only override one of the URLs. Confirm that you get errors about the others missing.
  - [x] Specify `BACKEND=stage` and try to override a URL. Confirm that the env variable is ignored, because overriding is all-or-nothing.

## Routes Added/Updated
None


README.md
```diff
@@ -54,65 +71,88 @@ uri is correct.  If it needs a trailing slash, be sure to include a trailing sla
...
+Because of the potential for this file to include sensitive information, **we strongly recommend adding this file to 
+`.gitignore`** for your project.
 
-#### Using the Test API
+#### Alternate option: Environment variables
+If you do not wish to use file-based configuration, any of the settings above can be overridden individually as 
+environment variables with the same name as appears in the config file. Eg
 
-To do this, you will need to [create a developer application](https://test.osf.io/settings/applications/) on the relevant version of the OSF.
+`BACKEND=test CLIENT_ID=gibberish ember serve`
```
abought: Reminder to self: this should be `ember server` for consistency with ember-cli docs. 

Very trivial, so I'll bundle in that along with any other changes required during code review.

index.js
```diff
@@ -6,96 +6,76 @@ var Funnel = require('broccoli-funnel');
 var mergeTrees = require('broccoli-merge-trees');
 var compileSass = require('broccoli-sass-source-maps');
 
+// Fetch a list of known backends. The user can always choose to override any of these URLs via ENV vars
+var knownBackends = require('./config/backends');
```
abought: That's a relief; I've been burned by ES6 in node4 before, since `index.js` doesn't appear to get transpiled. Huzzah for quality of life!

README.md
brianjgeiger: @abought Did you want to change this?

abought: Changed in ce6fc30

It doesn't affect functionality, but I tend to prefer the "official" forms of commands when I write docs. It's a bit easier for beginners if they don't have to look up aliases.

PR #172, opened 2017-01-27
Feature/EOSF-466

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/EOSF-466

# Purpose
Copyright date in footer needs to be updated from 2016 to 2017

# Notes for Reviewers
I used component from ember-osf instead of creating a new one. I updated the ember-osf component so that it returns the current year instead of having to update manually every year.
This required changes in ember-osf.


tests/integration/components/osf-copyright/component-test.js
```diff
@@ -2,25 +2,13 @@ import {
     moduleForComponent,
     test
 } from 'ember-qunit';
-import hbs from 'htmlbars-inline-precompile';
+
```
abought: With the removal of this code, what exactly does this unit test do?

I'm glad you're cleaning up unused code, but if I read this correctly, we should either:
- Write a test that does something, or,
- Remove this file completely (because it makes the test suite run longer without providing any information)

Long term I lean towards option A, but I'm not sure when we'll start emphasizing tests. Ember-osf test coverage has historically been a bit low.

addon/components/osf-copyright/component.js
```diff
@@ -11,5 +11,9 @@ import layout from './template';
  * @class osf-copyright
  */
 export default Ember.Component.extend({
-    layout
+    layout,
+    currentYear: function() {
+        let date = new Date();
```
abought: Unit testing tip: for the most part, this will be a straightforward test to write.

Our addons use a tool called Sinon.js to "stub" out things whose behavior we want to control- eg we wouldn't want to have to update the date unit test every year!

These links might be helpful (there's lots of stackoverflow posts of good info also):
http://sinonjs.org/docs/#clock
https://github.com/csantero/ember-sinon

My advice would be to pick a year you find interesting and make sure the component renders that year correctly. Bonus points if I learn something new about that year, but not required. ;)

samanehsan:
@abought, @sheriefvt is this good to merge? It would be great to have the correct copyright date on the UCSD dashboard (which is using the osf-copyright component) by Monday's demo.

abought:
According to my review, it's approved.  @brianjgeiger has a bit of a backlog of approved PRs to get through.

brianjgeiger:
Merged. Just pin to the commit above.

samanehsan:
Thanks @brianjgeiger! :octocat:

PR #175, opened 2017-02-02
[random] Add code coverage reporting [#EOSF-482]

abought:
## Ticket
https://openscience.atlassian.net/browse/EOSF-482

# Purpose
Add automatic code coverage reporting to ember-osf, and integrate with the `coveralls.io` external service.

This helps find exactly what parts of the code are in need of testing, line-by-line. See a sample report for one particular file:

https://coveralls.io/builds/9917692/source?filename=app%2Ftemplates%2Findex.hbs

# Notes for Reviewers

- Once conceptually approved, this will require a coordinated change to turn this repo on in coveralls.io ( @abought can flip the switch)

- You can run reports locally via `yarn run test:cover`, and reviewing the output report at `coverage/index.html`. Once coveralls integration is turned on, reports will be generated after every PR and viewable online by anyone.

- Presently this does not exclude the dummy app for addon tests, which may be a known issue. See `https://github.com/kategengler/ember-cli-code-coverage/issues/41`

## Routes Added/Updated
- None.


brianjgeiger:
I like this is principle. I know static analysis SAAS sites tend to go away frequently. This one may be simple enough that it won't, but please document any thoughts on likelihood of this place being around for a while and the feasibility or infeasibility of using a tool instead of a site.

The worst that will happen is it goes away and we no longer have coverage stats, which isn't the worst thing, of course, but I am curious.

abought:
> The worst that will happen is it goes away and we no longer have coverage stats, which isn't the worst thing, of course, but I am curious.

This is a good question- depending on external services is always a leap of faith (be it travis, coveralls, or even github). Coveralls has been around a while and is widely used.

That said, you will always be able to run the tooling in this command locally via `yarn run test:cover` (I'd be happy to show you the local HTML report). The basic coverage tools are open source as well; if coveralls went away, we'd really just be losing the badge.

The key is that coverage is run locally and produces a standard output format. The coveralls integration just comes at the end, when travis uploads the output to their server.

abought:
Updated to resolve merge conflicts.

brianjgeiger:
@abought When you have a moment, please do the post-deployment items

abought:
Activated on coveralls.io!

I've  been noticing some weirdness on my personal repo (demo-ember-osf), so we'll monitor this in case further tweaks are needed.

PR #176, opened 2017-02-02
Component for improving pagination results [EOSF-520]

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/EOSF-520

# Purpose
Improved pagination for preprints search results. Use a 3rd party component on preprints. The component can be used by other application (e.g. share). 
Currently, there are different implementations of pagination in ember-osf that might be redundant. Add a deprecation warning message to existing pagination components.

# Notes for Reviewers


addon/components/osf-paginator/component.js
```diff
@@ -24,6 +27,13 @@ import layout from './template';
 export default Ember.Component.extend({
     layout,
     currentPage: 1,
+    init() {
+        this._super(...arguments);
+        Ember.deprecate('osf-paginator is now deprecated. Use pagination-pager instead', false, {
+            id: 'osf-paginator',
+            until: '0.1.0'
```
abought: Thanks for adding this!  This is a good habit to get into as we start removing old stuff.

I suggest changing this to `1.0.0` as follows.

From the deprecation docs, this parameter should be a future release, rather than specifying the current version number:
> until String
The version of Ember when this deprecation warning will be removed.

We will be removing an entire component, which is not a small change! According to the [semver](http://semver.org/) specification that this addon follows, we should only make backwards-incompatible changes on a major version, eg `0.1.0 -> 1.0`.

As written, this warning implies that the component has already been removed ("will be gone by version 0.1"), which could be confusing.

addon/components/pagination-control/component.js
```diff
@@ -21,6 +24,14 @@ export default Ember.Component.extend({
         return this.get('currentPage') >= this.get('pageCount');
     }),
 
+    init() {
+        this._super(...arguments);
+        Ember.deprecate('pagination-control is now deprecated. Use pagination-pager instead', false, {
+            id: 'pagination-control',
+            until: '0.1.0'
```
abought: See comment above. Should also be 1.0.

sheriefvt:
Completed requested changes..thanks for your feedback and comments.

yarn.lock
```diff
@@ -4420,6 +4420,10 @@ output-file-sync@^1.1.0:
     mkdirp "^0.5.1"
     object-assign "^4.1.0"
 
+pagination-pager@2.4.2:
```
abought: I notice that you don't add any dependencies to package.json in this PR- was committing the change to your yarn.lock file intentional?

sheriefvt: i will reset this..sorry

sheriefvt:
did the change thanks

abought:
So: 

I really like that you updated the docstrings in the component, and I think it's really good that you inspected them locally to make sure they looked right before moving the PR forward.

But, er- could I ask you to revert just commit 41ea208 ? We plan to update the compiled HTML docs when a new release comes out, but doing it in every PR along the way adds 2300 lines of noisy commits to the diff, and creates more potential for merge conflicts later.

https://www.atlassian.com/git/tutorials/undoing-changes#git-revert

I'm very sorry for the inconvenience; I should have been more clear. We're not in the habit of updating our docs yet so this process is still a bit rough. :)

PR #182, opened 2017-02-13
[EOSF-512] Add blogurl as a preprint provider field

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-512


brianjgeiger:
When you have a moment please resolve conflicts.

abought:
Possible dupe: https://github.com/CenterForOpenScience/ember-osf/pull/193

Because `fixstring` alters the text (to clean up user-entered values), it could in theory subtly change the link destination. Once API blockers are resolved, I recommend choosing which approach we want and closing whichever of the two tickets we don't want.

laurenbarker:
Closing since the related ticket is closed.

PR #188, opened 2017-02-28
Feature/New OSF Navbar [EOSF-511]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-511

Companion PR ❗️ Styles have been moved to osf-style guide https://github.com/CenterForOpenScience/osf-style/pull/95

# Purpose
Ember-OSF component that is a redesign of the navbar. For use in preprints and registries.  Ensures that navigation within services is easy and clear without needing two stacked navigation bars, so that visitors to the site can easily orient themselves.

# Screenshots 
### Home primary navigation selected
screen shot 2017-04-03 at 2 54 11 pm

### Preprints primary navigation selected
screen shot 2017-04-03 at 2 55 05 pm

### Service dropdown
screen shot 2017-04-03 at 3 06 09 pm

### Auth dropdown
screen shot 2017-04-03 at 3 06 15 pm


addon/components/new-osf-navbar/style.scss
```diff
@@ -0,0 +1,228 @@
...
+            li > a {
+               font-size: 15px;
+            }
+            top: 60px;
+            left: 15px;
+        }
+    }
+
+    /* Overrides background of dropdown-toggle when active - needed for auth navbar */
+     .navbar-nav > .open > a,
```
abought: Will this navbar be used in the regular OSF too?

Currently, navbar CSS lives in the [osf-style](https://github.com/CenterForOpenScience/osf-style) repo. For the sake of maintainability, I think we should consider putting the styles for this shared product in that shared location. Then we would maintain a single location for visual elements of both the flask and ember apps.

pattisdr: Yes, thanks, this is commented in the ticket.  Before I moved this over to osf-style I wanted to confirm that there are not other navbars in the OSF besides the OSF-navbar that are using these navbar classes.

abought:
For the home navbar in particular: the shape, color difference, and rightmost "SEARCH" label all serve as common visual indicators of a search field. However, this seems to be just a rectangle, not a textbox that allows search.

Is there anything that we can do to visually differentiate the secondary nav area, without making it look like an interactive UI element in itself?

If you haven't had an opportunity to chat with @caneruguz about this yet, I've often found great benefit from consulting his UX expertise.

brianjgeiger:
Look and feel issues go through Product. If they're happy with it, then they'll get the chance to try it out and see how the users feel about it when it goes live, but I'd like to reduce second-guessing on that front.  In this case, I believe Sonya did the design.

pattisdr:
Thanks for your input. The secondary navigation area is not intended to be interactive UI element, it's just a navbar. In my opinion, this doesn't look like a search field, because there are other elements in the navigation area, like the current service "HOME", and the "My Projects" link.  It also doesn't behave like a search field - only the links are clickable. When you click on "Search" in the Home navbar, the green search bar that we use will drop down.   

I built this referencing the wireframes in the navbar spec (see link in ticket), drawn by Sonya, I believe.  I can pass on your concerns to Product.

abought:
Here is a sample of a few navbars from other sites, for comparison. Note the common usage of shape, color contrast, the word "search", and positioning at the top of the page as cues for a search field. It doesn't act like a search field, but I spent a good 30 seconds trying to click it all the same...

I think the secondary nav area is a good idea, but it might help avoid confusion if we tone down the visual cues a bit. Any good design starts out as a wireframe, and is iterated once it gets in front of users.

Github:
screen shot 2017-03-01 at 1 44 35 pm

Tumblr:
screen shot 2017-03-01 at 1 46 11 pm

MDN:
screen shot 2017-03-01 at 1 49 49 pm

See also facebook, etc.

brianjgeiger:
@abought I appreciate the input, but please don't use code review for design comments.

addon/components/search-dropdown/template.hbs
```diff
@@ -3,7 +3,7 @@
         <div class="row">
             <div class="col-md-12">
                 <form class="input-group" {{action "search" on="submit"}}>
-                    <input id="searchPageFullBar" onchange={{action "updateQuery" value="target.value"}} name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search">
+                    <input id="searchPageFullBar" onchange={{action "updateQuery" value="target.value"}} name="search-placeholder" type="text" class="osf-search-input form-control" placeholder="Search the OSF">
```
abought: It appears that `updateQuery` purely exists to set a variable to the field value whenever the field changes.

This sounds like a use case for two-way data binding, which is [set up automatically](https://guides.emberjs.com/v2.6.0/templates/input-helpers/) (no action needed) if you use the curly brace helper form:

`{{input value=query}}`

Since `query` seems to be just a placeholder value, a two-way binding might be a concise approach for this specific situation.

addon/helpers/build-secondary-nav-links.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+export function buildSecondaryNavLinks(params/*, hash*/) {
+    const currentService = params[0].toUpperCase();
+    const session = params[1];
+    const osfUrl = config.OSF.url;
+    let links = Ember.Object.create({
+        HOME: [
+            {
+                name: `${session.isAuthenticated ? 'My Projects' : 'Browse'}`,
+                href: `${session.isAuthenticated ? `${osfUrl}myprojects/` : `${osfUrl}explore/activity/`}`
```
abought: Perhaps not now, but for the future: I'm starting to notice a number of places throughout the codebase that use service links. (eg navbar, footer) 

In the future we might want to consider centralizing those URL definitions in one place, especially for services that don't depend on whether or not the user is logged in.

addon/helpers/build-secondary-nav-links.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @module ember-osf
+ * @submodule helpers
+ */
+
+/**
+ * Returns secondary links corresponding to a given service
+ *
+ * @class buildSecondaryNavLinks
+ * @param {String} currentService name of current service (HOME, PREPRINTS, REGISTRIES, or MEETINGS)
```
abought: Rather than hardcode these constants with special meaning throughout the code, might I suggest defining them as predefined constants in a single shared file that everything else references? It might ease maintainability later.

addon/helpers/build-secondary-nav-links.js
```diff
@@ -0,0 +1,83 @@
...
+ * @return {Array} Returns array of secondary link information
+ */
+export function buildSecondaryNavLinks(params/*, hash*/) {
+    const currentService = params[0].toUpperCase();
+    const session = params[1];
+    const osfUrl = config.OSF.url;
+    let links = Ember.Object.create({
+        HOME: [
+            {
+                name: `${session.isAuthenticated ? 'My Projects' : 'Browse'}`,
```
abought: `session.isAuthenticated` [appears to be](https://github.com/simplabs/ember-simple-auth/blob/master/addon/services/session.js) a computed property. Would it make more sense to fetch it as `session.get('isAuthenticated')`, to ensure that we are getting the [calculated result](http://emberjs.com/api/classes/Ember.Service.html#method_get) (rather than just the existence) of the property?

addon/components/new-osf-navbar/style.scss
abought: Thanks. It appears that at present the consuming app has to install `osf-style` itself; when that time comes, this may take some coordination across consuming repos to update `osf-style`.
https://github.com/abought/demo-ember-osf/blob/develop/ember-cli-build.js#L49
https://github.com/abought/demo-ember-osf/blob/develop/bower.json

(it's definitely one of the pain points of using the addon at present- styles are pinned to a specific version, and the dependency version is defined independently in many different apps)

addon/components/new-osf-navbar/template.hbs
```diff
@@ -0,0 +1,75 @@
+<div class="new-osf-navbar-styling">
+    <nav class="navbar navbar-inverse" id="navbarScope">
+        <div class="container">
+            <div class="navbar-header">
+                {{!Toggle navigation button for xs screens}}
+                <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target=".secondary-navigation" {{action 'closeSearch'}}>
+                    <span class="sr-only">Toggle navigation</span>
```
abought: In general, we've been encouraging all ember apps to use `ember-i18n` out of the box: it's easier to build that in up front than to add it later.

Would it be considered within scope to request translation support before this PR goes forward?

addon/components/new-osf-navbar/template.hbs
```diff
@@ -0,0 +1,75 @@
...
+                    <span class="icon-bar"></span>
```
abought: To a user who can see, this will show an icon with three lines. But to a blind user with a screen reader, there is no indication what this element does. Might I suggest adding aria label text? Here is a general reference for another icon library that may serve as inspiration:
http://fontawesome.io/accessibility/

addon/components/new-osf-navbar/template.hbs
```diff
@@ -0,0 +1,75 @@
...
+                    <span class="icon-bar"></span>
+                    <span class="icon-bar"></span>
+                </button>
+                {{!OSF Brand}}
+                <a class="navbar-brand" href="/"><span class="osf-navbar-logo"></span> OSF </a>
+                {{!Primary navigation - toggle between OSF, Preprints, Registries, and Meetings}}
+                <ul class="primary-nav">
+                    <li class="dropdown">
+                        <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false" onclick={{action 'closeSecondaryAndSearch'}}><span class="caret"></span></a>
```
abought: See notes about semantic markup elsewhere- can this be a `<button>`?

abought: How does this UI element icon present to a blind user? Should there be an ARIA label? (I can help run this page with a screenreader if you'd like to test this out)

addon/components/new-osf-navbar/template.hbs
```diff
@@ -0,0 +1,75 @@
...
+            <div class="navbar-header">
+                {{!Toggle navigation button for xs screens}}
+                <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target=".secondary-navigation" {{action 'closeSearch'}}>
+                    <span class="sr-only">Toggle navigation</span>
+                    <span class="icon-bar"></span>
+                    <span class="icon-bar"></span>
+                    <span class="icon-bar"></span>
+                </button>
+                {{!OSF Brand}}
+                <a class="navbar-brand" href="/"><span class="osf-navbar-logo"></span> OSF </a>
```
abought: This is a clickable link with no text. Perhaps an ARIA label might be helpful?

addon/components/new-osf-navbar/template.hbs
pattisdr: I thought the "sr-only" class did a similar thing.  I'll remove it and add an aria-label.

abought: Actually, for this case here you make a fair comment- I may have gone through and tagged bare icons en masse. ARIA labels are the default mode for accessibility where `sr-only` is more bootstrap specific.

I'll add this link here for this specific case and let you sort out which is most appropriate for this line (answer may not be an "or"). Am curious how this turns out!
http://stackoverflow.com/questions/39223952/accessibility-sr-only-or-aria-label

addon/components/new-osf-navbar/template.hbs
```diff
@@ -0,0 +1,75 @@
...
+                {{!Primary navigation - toggle between OSF, Preprints, Registries, and Meetings}}
+                <ul class="primary-nav">
+                    <li class="dropdown">
+                        <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false" onclick={{action 'closeSecondaryAndSearch'}}><span class="caret"></span></a>
+                        <ul class="dropdown-menu service-dropdown" role="menu">
+                            {{#each osfServices as |service|}}
+                                <li><a href="{{service.url}}" onclick={{action "switchService" service.name}}>OSF<b>{{service.name}}</b></a></li>
```
pattisdr: That spec does not have a space.  My guess is that it is supposed to resemble the branding on preprints and registries.

tests/dummy/app/templates/application.hbs
```diff
@@ -1,4 +1,4 @@
-{{osf-navbar loginAction=(action 'login')}}
+{{new-osf-navbar loginAction=(action 'login') currentService='PREPRINTS'}}
```
pattisdr: Replaced old navbar with new navbar. I wasn't sure if we wanted to keep it, and it was a good reference while I was building the new one. 

Second, I haven't come up with a good idea to identify the consuming service, but don't know if it's needed right now.  We'll just be using this in two places:
{{osf-navbar loginAction=(action 'login') currentService='PREPRINTS'}}
{{osf-navbar loginAction=(action 'login') currentService='REGISTRIES'}}.

I've heard that we might be moving preprints and registries into one repo which would probably change how the service needs to be detected as well.

pattisdr: @sheriefvt is working on a helper that reads from the config file in https://openscience.atlassian.net/browse/EOSF-553

addon/components/new-osf-navbar/style.scss
```diff
@@ -0,0 +1,228 @@
...
+     /* Overrides font-size and position of user name / gravatar */
+     a.nav-user-dropdown {
+         padding-right: 0;
+         font-size: 15px;
+     }
+
+    /* Overrides color of dropdown list hover (applies to both service and auth dropdowns) */
+    .dropdown-menu > li > a:hover,
+    .dropdown-menu > li > a:focus {
+          background-image: -webkit-linear-gradient(top, $osf-highlight-navbar 0%, $osf-highlight-navbar 100%);
```
binoculars: There's an [ember-cli addon](https://github.com/kimroen/ember-cli-autoprefixer) that helps with this. The benefit is that over time, we can eliminate the prefixes for unsupported browsers and not have to think about it.

addon/helpers/build-secondary-nav-links.js
pattisdr: The currentService that is passed in is now pulled from the hostAppName.

addon/components/osf-navbar/component.js
```diff
@@ -8,35 +11,62 @@ import config from 'ember-get-config';
...
-
+    osfServices: Ember.computed(function() {
+        return osfServices;
+    }),
+    serviceLinks: Ember.computed(function() {
+        return serviceLinks;
+    }),
     host: config.OSF.url,
-
+    currentService: Ember.computed(function() { // Pulls current service name from consuming service's config file
```
pattisdr: Good point about the ember apps merging.

addon/const/osf-services.js
```diff
@@ -0,0 +1,38 @@
...
+        name: 'REGISTRIES',
+        url: serviceLinks.registriesHome
+    },
+    {
+        name: 'MEETINGS',
+        url: serviceLinks.meetingsHome
+    }
+];
+
+export default osfServices;
```
pattisdr: I don't think they should be combined. osf-services is mainly going to be used on this new navbar.  However, the service links can be used everywhere there is a link.  osf-services needs service-links, but not the other way around.

pattisdr:
Changes made @abought.

pattisdr:
Closing in favor of https://github.com/CenterForOpenScience/ember-osf/pull/204

PR #189, opened 2017-03-03
[EOSF-509] Update factories and have factories be merged to consuming apps

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-509

# Purpose
Move factories to where they get merged with consuming apps, to be used on their tests.
Create/Update factories


test-support/factories/citation.js
```diff
@@ -0,0 +1,8 @@
+import FactoryGuy from 'ember-data-factory-guy';
+import faker from 'faker';
+
+FactoryGuy.define('citation', {
+    default: {
+        citation: faker.lorem.words(10)
```
abought: Remind me, if you please: does this set up a single value for every instance of the factory? I notice that in many of our other factories, we provide a function that can be called per-instance instead:

`() => faker.lorem.words(10)`

Noted for this one field, but if necessary we should review all fields in the PR that use `faker` and treat them consistently.

hmoco:
@abought, a lot of these factories are indeed theoretical code, but a lot of them are also being used in the preprints' tests, in this sister PR: https://github.com/CenterForOpenScience/ember-preprints/pull/293. Will come back to address your other comments

abought:
The requested changes look good and will improve on what we have. 

The best test of factories will be to start using them; ok to move forward from me!

PR #191, opened 2017-03-07
[EOSF-461] Update node and bower dependencies

binoculars:
## Ticket
https://openscience.atlassian.net/browse/EOSF-461

# Purpose

* Keeping things up to date is important, see http://emberjs.com/blog/2016/10/17/ember-2-9-released.html

# Notes for Reviewers

* Skipped some tests, but will throw a ticket in the backlog to fix these. They appear to be low priority. Assessing that updating these dependencies is more important right now.
* We should look at the dummy app and consuming apps like preprints to make sure there aren't any downstream errors affected by these changes.
* Some packages were not upgraded, because of breaking changes


addon/components/file-browser-item/template.hbs
```diff
@@ -1,3 +1,3 @@
 {{! TODO: configurable columns }}
 {{file-browser-icon item=item}}
-<a {{action 'open'}}>{{item.itemName}}</a>
+<a {{action 'open'}} role="link">{{item.itemName}}</a>
```
abought: Side note: I was under the impression that ARIA naturally provided certain basic roles if you used the right markup.

So for example, an `<a>` would already be understood to be a link. Is that not the case? I'd be curious to know and would be willing to help get started on a screen reader to try it out.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_link_role
> Note: Where possible, it is recommended that you use a native <a> elements rather than the link role, as native elements are more widely supported by older user agents and assistive technology. Native <a> elements also support keyboard and focus requirements by default, without need for additional customization.

http://w3c.github.io/aria-in-html/#rule1
> If you can use a native HTML element [HTML51] or attribute with the semantics and behaviour you require already built in, instead of re-purposing an element and adding an ARIA role, state or property to make it accessible, then do so.

abought: Ah, found the spec: https://w3c.github.io/html-aam/#html-element-role-mappings

binoculars: It should, but since there's no href, the linter is unhappy

tests/dummy/app/components/file-tree/template.hbs
```diff
@@ -1,12 +1,11 @@
 <div class="file-tree">
     {{#if onClickFile}}
-        <a {{action onClickFile root}}>{{root.name}}</a>
+        <a {{action onClickFile root}} role="link">{{root.name}}</a>
```
abought: Similar to above: links for navigation, button for interactive elements that keep you on the same page.

(we may not use this widget for much, so not a super high priority. But people tend to code by example, so there's that)

brianjgeiger:
Merge conflicts

brianjgeiger:
Merge conflicts, and move jira ticket to passed code review when it's done (and ping me so I can merge)

binoculars:
@brianjgeiger merge conflicts are gone. Tests passing, coveralls a little unhappy about skipping test

PR #194, opened 2017-03-16
Helper for Markdown Rendering [EOSF-562]

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/EOSF-562
https://openscience.atlassian.net/browse/EOSF-437

# Purpose
Markdown needs to be rendered in different places at the frontend (e.g. abstracts in preprints) or other apps using ember-osf.  Because Preprint abstracts are also OSF Project descriptions, we should limit this to a subset of 

- markdown:
- Bold
- Italics
- Links 
- ordered lists
- unordered lists
- paragraphs / line breaks
- mathjax

A new helper is added that uses the markdown-it library. The helper has options to enable/disable various rendering rules to limit/allow markdowns. 

Tests have been added for the new helper.

# Notes for Reviewers


package.json
```diff
@@ -62,6 +62,7 @@
     "ember-infinity": "0.2.3",
     "ember-load-initializers": "0.5.1",
     "ember-lodash": "4.17.0",
+    "ember-markdown-it": "0.0.5",
```
pattisdr: Two things here:

1) You don't seem to be using this [addon](https://github.com/HeroicEric/ember-markdown-it), except that I suspect it's helping install markdown-it for you, because the markdown-it line in bower.json is not sufficient to make markdown-it available to the app.

2) Just to clarify, we are not using this addon, and making a custom helper that uses markdown-it instead, because we need to override specific things, like disabling images?

bower.json
```diff
@@ -14,7 +14,9 @@
     "toastr": "^2.1.2",
     "font-awesome": "~4.5.0",
     "typeahead.js": "^0.11.1",
-    "jquery.tagsinput": "^1.3.6"
+    "jquery.tagsinput": "^1.3.6",
+    "loaders.css": "^0.1.2",
+    "markdown-it": "^8.3.1"
```
pattisdr: If you are not planning on using the `ember-markdown-it` addon, I would remove it from `package.json`.  Then, you are going to need a few things to make markdown-it work - (this is how I got it working - there may be a better way).

1) Add `app.import(path.join(app.bowerDirectory, 'markdown-it/dist/markdown-it.js'));` to `ember-cli-build.js`

2) You'll now see something like this: `Could not find module: markdown-it`.  Remove `import markdownit from 'markdown-it';` from `addon/helpers/markdown-viewer.js`.   It looks like markdownit is available as a global.

3) jshint ignore:line where markdown it is used in `addon/helpers/markdown-viewer.js`

bower.json
```diff
@@ -14,7 +14,9 @@
     "toastr": "^2.1.2",
     "font-awesome": "~4.5.0",
     "typeahead.js": "^0.11.1",
-    "jquery.tagsinput": "^1.3.6"
+    "jquery.tagsinput": "^1.3.6",
+    "loaders.css": "^0.1.2",
```
pattisdr: How are you using `loaders.css`?

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { htmlSafe } = Ember.String;
+
+export function markdownViewer(params) {
+
+    const [markdown] = params;
+    const html = markdownit({
+        html: false,        // Enable HTML tags in source
+
+        linkify: false,     // Autoconvert URL-like text to links
```
pattisdr: Make true by default? I wasn't in any of the conversations about this helper, but from the PR description, looks like we want to allow links?

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,33 @@
...
+        // Double + single quotes replacement pairs, when typographer enabled,
+        // and smartquotes on. Could be either a String or an Array.
+        //
+        // For example, you can use '«»„“' for Russian, '„“‚‘' for German,
+        // and ['«\xA0', '\xA0»', '‹\xA0', '\xA0›'] for French (including nbsp).
+
+        xhtmlOut: false,    // Use '/' to close single tags (<br />).
+        // This is only for full CommonMark compatibility.
+
+        breaks: false,      // Convert '\n' in paragraphs into <br>
```
pattisdr: Maybe make true by default, since we're allowing line breaks, according to PR description?

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,33 @@
...
+    }).disable(['image', 'link']);
```
pattisdr: Why are we disabling 'link' here?

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,33 @@
...
+    const renderedHtml = html.render(markdown);
+
+    return htmlSafe(renderedHtml);
```
pattisdr: I've been told not to use this [before](https://github.com/CenterForOpenScience/ember-osf/pull/169#discussion_r110465172).  Can we be sure this is safe here?  It looks like HTML is disabled in this helper so I think that takes care of a lot it. But are there any other things we should do to ensure this is safe?

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,33 @@
+import Ember from 'ember';
+import markdownit from 'markdown-it';
+
+const { htmlSafe } = Ember.String;
+
+export function markdownViewer(params) {
```
pattisdr: Looking at this [comment](https://openscience.atlassian.net/browse/PRODUCT-242?focusedCommentId=45402&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45402), it seems like this helper should limit which markdown capabilities are available. Headers, for example, are still allowed. Code, is still allowed.

sheriefvt:
Thanks @pattisdr, will work on the requests changes.

addon/helpers/markdown-viewer.js
pattisdr: Also, can we get some yuidoc documentation on this helper?

sheriefvt: yes sure

sheriefvt:
@pattisdr thanks for your valuable feedback. This is the summary of changes:

- Add yuidoc to the new helper
- Remove ember-markdown-it
- Use global markdownit to make helper function pass the linter 
- Restrict the permitted markdown to the set of items in the PR description
- Add new tests 
🐼

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,44 @@
...
+ ```
+
+ @class markdown-viewer-helper
+ @uses markdownit
+ */
+/* global markdownit */
+export function markdownViewer(params) {
+
+    const [markdown] = params;
+    const html = markdownit('zero', {
```
pattisdr: Looks like you're using "zero" to disable all rules, and then enable only a select few. If so, are L21-37 doing anything? If not, they can be removed.

addon/helpers/markdown-viewer.js
```diff
@@ -0,0 +1,44 @@
...
+        // and smartquotes on. Could be either a String or an Array.
+        //
+        // For example, you can use '«»„“' for Russian, '„“‚‘' for German,
+        // and ['«\xA0', '\xA0»', '‹\xA0', '\xA0›'] for French (including nbsp).
+
+        xhtmlOut: false,    // Use '/' to close single tags (<br />).
+        // This is only for full CommonMark compatibility.
+
+        breaks: true,      // Convert '\n' in paragraphs into <br>
+    }).enable(['list', 'newline', 'emphasis', 'paragraph']);
```
pattisdr: Add link here?  According to the PR description, we still want to enable links.

sheriefvt:
@pattisdr thanks for feedback I updated code with the changes 🐼

pattisdr:
Looks good to me 🌤

laurenbarker:
Closing since this the use case for this helper has been resolved in another way.

PR #195, opened 2017-03-20
Detecting Consuming Application in Ember-OSF [EOSF-553]

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/EOSF-553

# Purpose
Sometimes in Ember-OSF, we need to know the consuming application - to handle things differently for preprints or registries, for example. Right now we pass in a string, like "PREPRINTS", to the ember-osf component. 

{{ osf-navbar signupUrl=theme.signupUrl loginAction=(action 'login') currentService='PREPRINTS' }}

A new utility and helper (with unit tests) are added to ember-osf, to allow retrieving host application name from config. The helper is using the utility function.

The host app name can be passed to the component now through the template handlebar
{{ osf-navbar signupUrl=theme.signupUrl loginAction=(action 'login') currentService=(host-service-name)}}

or can be retrieved at the component controller by importing the utility function.

# Notes for Reviewers

### DIFF
--- a/addon/helpers/host-service-name.js
+++ b/addon/helpers/host-service-name.js
@@ -1,0 +1,18 @@
+import Ember from 'ember';
+import hostServiceName from '../utils/host-service-name';
+
+/**
+ Use the `host-service-name` utility function to retrieve the host application name.
+ Usage example:
+ ```handlebars
+ This is text we want to do: {{host-service-name}}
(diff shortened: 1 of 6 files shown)


sheriefvt:
The pull request will be closed and replaced with another solution that use mixins.

PR #199, opened 2017-04-07
[EOSF-95] Remove unneeded comment serializer logic

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-95

# Purpose
Logic in the comment serializer is no longer needed (I actually think it was preventing it from working properly)


binoculars:
Looks fine, but will probably cause lint errors. Can you remove unnecessary params and comments and use const?

PR #200, opened 2017-04-11
Fix typo in docs for search-dropdown component

atelic:
## Ticket
N/A
# Purpose
Make documentation better. This component doesn't work if you use it the way the sample suggests.

# Notes for Reviewers
😄


abought:
Thanks for the catch! I've opened a PR that fixes the original file, so will close this one but send all praise your way.

PR #201, opened 2017-04-13
[EOSF-601] Fix overabundance of autosave action firing on license widget

hmoco:
Also wants: https://github.com/CenterForOpenScience/ember-preprints/pull/341

## Ticket
https://openscience.atlassian.net/browse/EOSF-601

## Purpose
If autosave was turned on, the license widget would fire the autosave function anytime the observer noticed a change on any of the license fields. This led to the save method attached to that action triggering any time a user would type a single key. To prevent that, autosave was wrapped in a `Ember.run.debounce` function. However, that did not seemed to prevent any of the triggers, maybe due to how each run was being created (each keystroke might be in a separate Ember run which would cause the debounce function to not do anything, as run function are limited to acting in a singular run). This adds lodash debouncing to autosave, which is based on a timeout, and allows the function to be called on the trailing end of the debouncing, making sure that all changes are actually saved. The timer is currently set so that the function can only be called twice each second, and will often only be called once per field being edited. 

### Before (note the 'Save Count' in the console on the right):
screen shot 2017-04-13 at 3 15 14 pm

### After:
screen shot 2017-04-13 at 3 13 44 pm

This could be further improved by tweaking with the debounce number, and making sure the function is not called on initial data load (as that does not represent a state different from what is saved but still triggers the observer)


addon/components/license-picker/component.js
```diff
@@ -12,6 +13,16 @@ export default Ember.Component.extend({
     showCopyrightHolders: true,
     showCategories: true,
     allowDismiss: false,
+    init() {
+        this._super(...arguments);
+        // Debouncing autosave prevents a request per keystroke, only sending it
+        // when the user is done typing (trailing=true), debounce timer can be tweaked.
+        this.set('debouncedAutosave', _.debounce(() => {
```
pattisdr: Confirmed that using lodash debounce drastically decreases the number of times 'save' is called. 

I do not think the 'save' action should be called on page load. In preprints, the 'editLicense' action, which is triggered here, is called as soon as the page loads, even though there has really been no change in state. Because we have analytics on the editLicense action, this editLicense count in Keen/GA will be falsely inflated, being triggered every time someone visits this page.

addon/components/license-picker/component.js
```diff
@@ -54,14 +49,26 @@ export default Ember.Component.extend({
...
-            this.set('nodeLicense', this.get('licensesAvailable.firstObject'));
+        if (!this.get('licenses')) {
+            this.get('store').query('license', { 'page[size]': 20 }).then(ret => {
+                this.set('licensesAvailable', ret);
+                if (!this.get('currentValues.licenseType.id')) { //if not resolved properly
+                    this.set('nodeLicense', this.get('licensesAvailable.firstObject'));
+                } else {
+                    this.set('nodeLicense', this.get('currentValues.licenseType'));
+                }
+                this.set('skipAutosave', false);
```
pattisdr: Maybe pull L54-60 into a function that can be called in both branches here?

hmoco:
@brianjgeiger, tests pass locally and seem to be unrelated to the change, would you mind rerunning?

brianjgeiger:
Done, but let's keep an eye on this. I don't want probabilistic failures on our ember apps. That will need to be squashed quickly.

hmoco:
@brianjgeiger, I've never seen this one before, but there's definitely another error that I see occur every now and then in both this and ember-preprints if I recall correctly. It goes something like ` Can't call 'set' on a destroyed object `. The small amount of research I've done on it seem to say that it is not normally a probabilistic error, but it seems to be as rerunning tests usually makes it go away. I've also had that error occur consistently, but it was something I was definitely doing wrong. If you are worried about probabilistic errors, remind me to bring it up in scrum tomorrow to see how many people have run into it and if anyone has any leads on how to prevent it.

addon/components/license-picker/component.js
```diff
@@ -1,6 +1,8 @@
 import Ember from 'ember';
+import _ from 'lodash';
 import layout from './template';
 
+
 export default Ember.Component.extend({
```
pattisdr: While you're here, could you add some documentation for this component that includes sample usage? Also it would be good if the component properties were explained, especially the licenses, currentValues and autosave properties.

hmoco:
Companion PR made here: https://github.com/CenterForOpenScience/ember-preprints/pull/341

pattisdr:
Looks great to me @hmoco! The documentation on this is helpful, and the debouncing is working nicely.

PR #202, opened 2017-04-13
Add children field to taxonomy model [EOSF-491]

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/EOSF-491

# Purpose
When hovering over top-level discipline in the index page, a tooltip shows an example of subdisciplines. Currently, the subdisciplines are hard-coded, causing the display of one or more subjects that are not used (allowed) by the provider.

This PR is part of the solution presented in PR [326](https://github.com/CenterForOpenScience/ember-preprints/pull/326)
This solution requires updating the OSF API and returning subject children with each request. 
# Notes for Reviewers


laurenbarker:
Closing since this improvement was resolved by removing tooltips on subjects.

PR #205, opened 2017-05-03
Feature/Discover-Page Widget Improvements [EOSF-642]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-642

# Purpose
The discover page widget (EOSF-521) is merged, and now I am bringing  https://github.com/CenterForOpenScience/ember-preprints/pull/302 and https://github.com/CenterForOpenScience/ember-osf-registries/pull/11 up to date. Due to some recent merges, there need to be some changes Ember-side to the discover-page widget before it is used to build the discover pages for preprints and registries.

# Summary of changes
- Move over test for analytics mixin from preprints -> ember-osf
- Move over test for search result from preprints - > ember-osf
- Import ember-template-compiler in ember-cli-build
- Construct example search routes differently in lucene search help modal
- `date_updated` field which was pulled over from preprints search results does not always exist for registries search results. For now, don't show field if it does not exist.
- Modify total-share-results component

# Testing notes
This should be tested as part of EOSF-557 and EOSF-558 - testing the discover-page widget in preprints and registries.


addon/components/search-help-modal/component.js
```diff
@@ -15,7 +15,7 @@ export default Ember.Component.extend({
     layout,
     isOpen: false,
     currentPath: Ember.computed(function() {
-        return Ember.getOwner(this).lookup('controller:application').currentPath;
+        return window.location.origin + window.location.pathname;
```
binoculars: I'd recommend against using a computed property and instead set it in one of the [lifecycle hooks](https://guides.emberjs.com/v2.8.0/components/the-component-lifecycle/). You also may want to use window.location.href, or prefer template strings instead of the + operator.

addon/components/search-result/component.js
```diff
@@ -124,11 +124,13 @@ export default Ember.Component.extend(Analytics, hostAppName, {
     // Determines whether tags in search results should be links - preprints and registries are not using tag filter right now
     tagsInQueryParams: Ember.computed('queryParams', function() {
         let foundTags = false;
-        this.get('queryParams').forEach(param => {
-            if (param === 'tags') {
-                foundTags = true;
-            }
-        });
+        if (this.get('queryParams')) {
```
binoculars: this could be simplified to:

```js
return (this.get('queryParams') || []).includes('tags');
```

tests/integration/components/search-result/component-test.js
```diff
@@ -1,25 +1,92 @@
...
 moduleForComponent('search-result', 'Integration | Component | search result', {
-  integration: true
+    integration: true,
+    beforeEach: function() {
+        let noop = () => {};
+        this.set('noop', noop);
+    }
 });
 
+let result = Ember.Object.extend({
```
binoculars: Can we move this into a test factory?

addon/components/search-result/template.hbs
```diff
@@ -33,7 +33,9 @@
             </div>
 
             {{!Last edited on on}}
-            <div class="m-t-sm"> <em> {{t "eosf.components.searchResult.lastEdited"}}: {{moment-format result.date_updated 'YYYY-MM-DD'}} </em> </div>
+            {{#if result.date_updated}}
```
binoculars: Shouldn't this be camelCased? Model or serializer might need to be updated?

pattisdr: There are SHARE results - this was adapted from [the preprints search results component](https://github.com/CenterForOpenScience/ember-preprints/blob/develop/app/templates/components/search-result.hbs#L28)

tests/integration/components/search-result/component-test.js
pattisdr: I might be misunderstanding what you mean, but I tried to make it a factory under test_support and was getting an error that it wasn't a model.

addon/components/search-result/template.hbs
binoculars: Oh right, ya we just get the SHARE results directly from an ajax call.

PR #208, opened 2017-05-12
[EOSF-659] Make fa-icon a button rather than an i inside a button

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-659

# Purpose

Fix inability to click x to remove active filter in Firefox

# Summary of changes

Made the fa-icon a button tag instead of the default i tag and changed containing tag to be a span.

# Testing notes

Need to check in all browsers. Tested in Chrome, FF, and Safari on Mac.

Note: In Firefox the click event is attached to the button, even when the button contains another object. This differs from Chrome and Safari (unsure about IE). Try https://jsfiddle.net/5pvLez34/ in various browsers.

### DIFF
--- a/addon/components/discover-page/style.scss
+++ b/addon/components/discover-page/style.scss
@@ -77,7 +77,7 @@
     }
 }
 
-button.removeActiveFilter {
+.removeActiveFilter button {
     border: none;
     outline: none;
     background-color: transparent;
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Tests failed in a weird way. Restarting to see if it was just a Thing.

PR #210, opened 2017-05-18
[EOSF-601] Reverse debouncing logic on license widget

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-601

# Purpose
Reverse the debounce logic. Autosave makes sense to be calling a save action anytime any change happens, leaving the behavior of debouncing to be handled by whatever controller is consuming the component. In the use case in preprints, no debouncing is necessary. 

# Changes
A lot of small improvements were removed for the purpose of reversing debonucing logic (which was causing a bunch of bugs). Maybe worth looking into this again in the future. Ticket tracking those improvements being reintroduced: https://openscience.atlassian.net/browse/EOSF-671

### DIFF
--- a/addon/components/license-picker/component.js
+++ b/addon/components/license-picker/component.js
@@ -41,52 +41,30 @@ export default Ember.Component.extend({
     showCopyrightHolders: true,
     showCategories: true,
     allowDismiss: false,
-    init() {
-        this._super(...arguments);
-        // Debouncing autosave prevents a request per keystroke, only sending it
-        // when the user is done typing (trailing=true), debounce timer can be tweaked.
-        this.set('debouncedAutosave', _.debounce(() => {
-            if (this.get('autosave')) {
-                this.get('actions.save').bind(this)();
-            }
-        }, 500, {trailing: true}));
-    },
     didReceiveAttrs() {
-        if (!this.get('currentValues.year') && !this.get('year')) {
-            let date = new Date();
-            this.set('year', String(date.getUTCFullYear()));
-        }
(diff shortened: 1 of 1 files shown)


brianjgeiger:
And this is going to have to be a hotfix, since we just released 0.5.0 yesterday.

addon/components/license-picker/component.js
```diff
@@ -41,52 +41,30 @@ export default Ember.Component.extend({
...
+    _setNodeLicense: Ember.observer('licensesAvailable', 'currentValues.licenseType', function() {
+        if (!this.get('currentValues.licenseType.id')) { //if not resolved properly
+            this.set('nodeLicense', this.get('licensesAvailable.firstObject'));
+        } else {
+            this.set('nodeLicense', this.get('currentValues.licenseType'));
         }
-        this.set('showYear', text.indexOf('{{year}}') !== -1);
-        this.set('showCopyrightHolders', text.indexOf('{{copyrightHolders}}') !== -1);
-    }),
-    yearRequired: Ember.computed('nodeLicense', function() {
```
jamescdavis: the template is still expecting this to be defined

addon/components/license-picker/component.js
```diff
@@ -41,52 +41,30 @@ export default Ember.Component.extend({
...
-        return this.get('nodeLicense.requiredFields') && this.get('nodeLicense.requiredFields').indexOf('year') !== -1;
-    }),
-    copyrightHoldersRequired: Ember.computed('nodeLicense', function() {
```
jamescdavis: this one too

PR #215, opened 2017-06-06
[EOSF-631] extract DOI utility and helper

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-631

# Purpose

Moved method for extracting a DOI from a string into a utility and wrapped it with a helper.

# Summary of changes

Added utils/extract-doi-from-string and helpers/extract-doi

# Testing notes

Internally used utility. Should be tested in interfaces that parse DOIs from strings (e.g. preprint submission).

### DIFF
--- a/tests/unit/utils/extract-doi-from-string-test.js
+++ b/tests/unit/utils/extract-doi-from-string-test.js
@@ -1,0 +1,25 @@
+import extractDoiFromString from 'dummy/utils/extract-doi-from-string';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | extract doi from string');
+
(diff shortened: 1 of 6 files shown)


tests/unit/utils/extract-doi-from-string-test.js
```diff
@@ -0,0 +1,25 @@
+import extractDoiFromString from 'dummy/utils/extract-doi-from-string';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | extract doi from string');
+
+test('return DOI when DOI found', function(assert) {
```
hmoco: 👍

PR #223, opened 2017-06-08
Feature/Discover Page Changes for ASU Repository [EOSF-680]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-680
❗️ is needed for https://github.com/CenterForOpenScience/ember-preprints/pull/383

# Purpose
Make changes to the discover-page widget that are needed for the ASU repository.

# Summary of changes
- Provider is passed into component
- Search-facet-worktype component is now taking care of building the list of types itself, instead of passing them into the component.  This allows preprints to be able to use this facet.
- Only show add preprint box on submissions page allowSubmissions is true.
- Conditionally modify search query.  If additionalProviders exist, pass all providers (like ASU, Symbiota, etc. into query). Otherwise, just pass single provider like engrxiv.
- Tags on search result are conditionally clickable (allowed if additionalProviders exist, because repositories with additionalProviders are using the tags facet)

# Testing notes

### DIFF
--- a/addon/components/discover-page/component.js
+++ b/addon/components/discover-page/component.js
@@ -48,6 +48,7 @@ import { getUniqueList, getSplitParams, encodeParams } from '../../utils/elastic
  *    showActiveFilters=showActiveFilters
  *    sortOptions=sortOptions
  *    subject=subject
+ *    themeProvider=themeProvider
 * }}
  * ```
  * @class discover-page
@@ -286,6 +287,12 @@ export default Ember.Component.extend(Analytics, hostAppName, {
      * @default ''
      */
     tags: '',
+    /**
+     * themeProvider
+     * @property {Object} Preprint provider loaded from theme servicer
+     * @default ''
+     */
+    themeProvider: null,
     took: 0,
     /**
      * type query parameter.  If "type" is one of your query params, it must be passed to the component so it can be reflected in the URL.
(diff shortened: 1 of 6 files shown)


addon/components/discover-page/component.js
```diff
@@ -348,11 +354,6 @@ export default Ember.Component.extend(Analytics, hostAppName, {
             this.setActiveFiltersAndReload('activeFilters.providers', filter.split('OR'));
         }
     })),
-    processedTypes: Ember.computed('types', function() {
```
hmoco: Hey! Looked over this for a while, and can't seem to figure out what they did in the first place. Could you provide some insight into why they were here and why they are no longer needed?

hmoco:
This looks good!

addon/components/discover-page/component.js
pattisdr: Hey Henrique - This was code copied completely from SHARE.   SHARE uses it to build the options in the type facet: 
screen shot 2017-06-13 at 10 27 46 am.  Instead of having it in the base discover-page widget, I moved it to the widget that is using it, the search-facet-worktype facet.  This is so ASU could use it.  How it was written before, you could only really use this widget if you were using all of the standard SHARE facets as well. But since ASU is only using a selection of these facets, where it was defining which facets it wanted to use, it didn't have access to these getTypes methods, etc. to build the types. So by letting the facet itself take care of it, we open up more options to how these widgets can be combined.

PR #225, opened 2017-06-16
[EOSF-554] Update to Node.js version 8

binoculars:
❗️Needs CenterForOpenScience/osf.io#7350

## Ticket
https://openscience.atlassian.net/browse/EOSF-554

# Purpose
Updates Node.js version to 8

# Summary of changes
Remove lts/boron and replace it with 8 everywhere, and add --ignore-engines due to  qunitjs/qunit#1184.

# Testing notes
This should not affect any functionality, normal regression only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:boron
+FROM node:8
 
 RUN apt-get update \
     && apt-get install -y \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:boron
+FROM 8
```
jamescdavis: Shouldn't this be `node:8`?

binoculars: I must've gotten a little paste happy

PR #231, opened 2017-06-26
[EOSF-696]Add style path to share logo on the provider's asset directory

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-696

# Purpose
Use the same path for both white and black share logos, witht he provider directory having the asset wanted by the provider.

### DIFF
--- a/addon/components/discover-page/template.hbs
+++ b/addon/components/discover-page/template.hbs
@@ -18,7 +18,7 @@
                         <h1> {{discoverHeader}} </h1>
                     {{/if}}
                     {{!POWERED BY SHARE}}
-                    <p class="lead">{{t 'eosf.components.discoverPage.poweredBy'}} <a href="https://share.osf.io/" id='share-logo' class="share-logo" title="SHARE" onclick={{action "click" "link" "Discover - SHARE Logo"}}></a></p>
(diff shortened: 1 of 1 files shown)


addon/components/discover-page/template.hbs
```diff
@@ -18,7 +18,7 @@
                         <h1> {{discoverHeader}} </h1>
                     {{/if}}
                     {{!POWERED BY SHARE}}
-                    <p class="lead">{{t 'eosf.components.discoverPage.poweredBy'}} <a href="https://share.osf.io/" id='share-logo' class="share-logo" title="SHARE" onclick={{action "click" "link" "Discover - SHARE Logo"}}></a></p>
+                    <p class="lead">{{t 'eosf.components.discoverPage.poweredBy'}} <a href="https://share.osf.io/" id='share-logo' style="background-image: url('./powered_by_share.png')" class="share-logo" title="SHARE" onclick={{action "click" "link" "Discover - SHARE Logo"}}></a></p>
```
brianjgeiger: Doesn't some CSS have to change as well?

brianjgeiger: Actually, I don't think this is right. It should be using the assets url generating component / helper thing.

brianjgeiger: Oooh, which could be a problem, couldn't it?

PR #234, opened 2017-06-30
Feature/remove shareTotal request for branded preprint providers [EOSF-720]

laurenrevere:
## Ticket
https://openscience.atlassian.net/browse/EOSF-720

# Purpose
Remove “X many preprints have been archived” from  Paleorxiv.

# Summary of changes
Now for branded preprint providers shareTotal is set to zero and there is no longer a request to grab the number of share results.

# Testing notes
Does Paleorxiv no longer show a “X many preprints have been archived.” Does it still show up on registries and the landing page.

### DIFF
--- a/addon/components/discover-page/template.hbs
+++ b/addon/components/discover-page/template.hbs
@@ -40,7 +40,9 @@
                 {{search-help-modal isOpen=showLuceneHelp}}
 
                 {{!NUMBER OF SEARCHABLE RESULTS}}
-                {{total-share-results}}
+                {{#if (not theme.isProvider)}}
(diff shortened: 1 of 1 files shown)


addon/components/total-share-results/component.js
```diff
@@ -68,23 +68,25 @@ export default Ember.Component.extend(hostAppName, {
         };
 
         if (this.get('theme.isProvider')) {
+            this.set('shareTotal', 0);
```
jamescdavis: Good idea to make sure shareTotal will always have a reasonable value.

PR #237, opened 2017-07-03
[EOSF-541] Shorten "Add a preprint" to "Add" in navbar for small screens

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-541

# Purpose

Shorten "Add a preprint" to "Add" for small and xs screens"

# Summary of changes

Added template logic to display shortened version for small and xs screens.

# Testing notes

Breakpoint is at 992px wide

### DIFF
--- a/addon/components/new-osf-navbar/template.hbs
+++ b/addon/components/new-osf-navbar/template.hbs
@@ -48,12 +48,12 @@
                             {{!SERVICE LINK}}
                             <li>
                                 <a href="{{navLink.href}}">
-                                    {{#if (eq navLink.name (t 'eosf.navbar.addAPreprint'))}}
+                                    {{#if (eq navLink.type 'addAPreprint')}}
                                         <span class="hidden-xs hidden-sm">
-                                            {{t 'eosf.navbar.add'}}
+                                            {{navLink.name}}
                                         </span>
                                         <span class="hidden-md hidden-lg hidden-xl">
-                                            {{navLink.name}}
+                                            {{t 'eosf.navbar.add'}}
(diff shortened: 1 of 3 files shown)


addon/components/new-osf-navbar/template.hbs
```diff
@@ -46,7 +46,20 @@
...
                         {{else}}
                             {{!SERVICE LINK}}
-                            <li><a href="{{navLink.href}}"> {{navLink.name}}</a></li>
+                            <li>
+                                <a href="{{navLink.href}}">
+                                    {{#if (eq navLink.type 'addAPreprint')}}
+                                        <span class="hidden-xs hidden-sm">
+                                            {{navLink.name}}
+                                        </span>
+                                        <span class="hidden-md hidden-lg hidden-xl">
```
hmoco: Seems like this could be simplified using `hidden-md-up`. The other case (4 lines above) could be simplified using `hidden-sm-down`: https://v4-alpha.getbootstrap.com/layout/responsive-utilities/.
Unless that is not available in our current version of bootstrap, which I suspect is the case :(

hmoco:
@jamescdavis, could you respond to the one comment one way or another so I can pass? This looks good otherwise!

addon/components/new-osf-navbar/template.hbs
jamescdavis: Yeah, I'm pretty sure those are new in v4. I think we just have http://getbootstrap.com/css/#responsive-utilities-classes available to us currently.

PR #238, opened 2017-07-12
[EOSF-754] Fix display of custom taxonomy subjects

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-754

# Purpose
Subjects are now being passed back from SHARE as strings contained their hierarchy. We want to show to the user the same things we used to (deduplicated parents and children as separate 'tag'-like subjects), but retain the hierarchy and taxonomy tag (i.e. 'bepress'), so they can be filtered properly.

# Summary of changes
Add custom taxonomy helper so filters can be shown properly

Add logic to separate out and deduplicate all subjects in a search result, to be displayed as such but also filtering using their original value

### Before:
screen shot 2017-07-12 at 2 29 57 pm

### After:
screen shot 2017-07-12 at 2 29 24 pm
screen shot 2017-07-12 at 2 29 33 pm

### DIFF
--- a/addon/components/search-result/component.js
+++ b/addon/components/search-result/component.js
@@ -90,12 +90,12 @@ export default Ember.Component.extend(Analytics, hostAppName, {
         let uniqueSubs = {}
         subs.map(e => {
             let tax, subjects;
-            [tax, ...subjects] = e.text.split('/');
+            [tax, ...subjects] = e.text.split('|');
             if (subjects.length) { //accounting for non-custom taxonomy subjects, if we ever get back to that
                 for (var i = 0; i < subjects.length; i++) {
                     uniqueSubs[subjects[i]] = {
                         text: subjects[i],
-                        value: [tax, ...subjects.slice(0, i+1)].join('/'),
+                        value: [tax, ...subjects.slice(0, i+1)].join('|'),
                         taxonomy: tax
                     }
                 }
--- a/addon/components/search-result/component.js
+++ b/addon/components/search-result/component.js
@@ -86,18 +86,25 @@ export default Ember.Component.extend(Analytics, hostAppName, {
         return (this.get('result.tags') || []).slice(0, this.get('maxTags'));
     }),
     subjects: Ember.computed('result.subjects', function() {
-        // TODO: how to check if this app actually uses custom taxonomies
         let subs = this.get('result.subjects');
         let uniqueSubs = {}
         subs.map(e => {
             let tax, subjects;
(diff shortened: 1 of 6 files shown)


addon/helpers/custom-taxonomy-filter.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
+
+var customTaxonomies = ['bepress'];
+
+export function customTaxonomyFilter(params/*, hash*/) {
+    let words = params[0].split('/');
+    if (customTaxonomies.indexOf(words[0]) !== -1 && words.length > 1) {
```
jamescdavis: I think, instead of looking for specific taxonomies, we'll just have to assume, if it has a slash in it, it's a custom taxonomy.

addon/components/search-result/component.js
```diff
@@ -86,7 +86,29 @@ export default Ember.Component.extend(Analytics, hostAppName, {
         return (this.get('result.tags') || []).slice(0, this.get('maxTags'));
     }),
     subjects: Ember.computed('result.subjects', function() {
-        return (this.get('result.subjects') || []).slice(0, this.get('maxSubjects'));
+        let subs = this.get('result.subjects');
+        let uniqueSubs = {}
+        subs.map(e => {
+            let tax, subjects;
+            [tax, ...subjects] = e.text.split('/');
+            if (subjects.length) { //accounting for non-custom taxonomy subjects, if we ever get back to that
```
jamescdavis: Probably a good idea, just in case.

addon/helpers/custom-taxonomy-filter.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
+
+var customTaxonomies = ['bepress'];
```
jamescdavis: Should drop this now.

tests/unit/helpers/custom-taxonomy-filter-test.js
```diff
@@ -0,0 +1,24 @@
+import { customTaxonomyFilter } from 'dummy/helpers/custom-taxonomy-filter';
+import { module, test } from 'qunit';
+
+module('Unit | Helper | custom taxonomy filter');
+
+test('replace very deep custom taxonomy hierarchy with just final subject name', function(assert) {
+    let result = customTaxonomyFilter(['bepress|Fruits|Bananas|Green Bananas|Small / Green Bananas|Small Green Bananas with Polka Dots']);
```
jamescdavis: 👍 for the subject w/ slash

PR #242, opened 2017-07-20
Update SHARE filter for custom taxonomies [EOSF-753]

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-753

# Purpose

Update SHARE subject filter to work with custom taxonomies.

# Summary of changes

Changed from terms filter to multiple should match filters.

# Testing notes

This only makes subject filters work when clicking on preprint tags on search results.

### DIFF
--- a/addon/components/discover-page/component.js
+++ b/addon/components/discover-page/component.js
@@ -468,15 +468,25 @@ export default Ember.Component.extend(Analytics, hostAppName, {
             const val = filterMap[key];
             const filterList = activeFilters[key];
             this.set(key, filterList.join('OR'));
-
             if (!filterList.length || (key === 'providers' && this.get('theme.isProvider')))
                 return;
 
-            filters.push({
-                terms: {
-                    [val]: filterList
-                }
-            });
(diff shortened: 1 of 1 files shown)


addon/components/discover-page/component.js
```diff
@@ -472,11 +472,22 @@ export default Ember.Component.extend(Analytics, hostAppName, {
             if (!filterList.length || (key === 'providers' && this.get('theme.isProvider')))
                 return;
 
-            filters.push({
-                terms: {
-                    [val]: filterList
+            if (val == 'subjects') {
```
hmoco: I might be missing something, but this looks like it would change the behavior not only on 'subjects', but if the filter type is something else, it would not push to 'filters'. Shouldn't the original `filters.push({terms: {[val]: filterList })` be on an else? Or is the behavior for things like provider filter handled elsewhere?

jamescdavis: You are totally correct. Thank you for catching this!

PR #255, opened 2017-08-10
Make add a preprint not use preprintWords

hmoco:
# Purpose
Fixes small bug that cant find preprint words to replace preprint in this instance due to how navbar access `t` helper


jamescdavis:
This was fixed by #256

PR #277, opened 2017-09-25
[Investigative][Do Not Merge] Feature/Django Waffle Exploration with Ember [OSF-8685]

pattisdr:
## Ticket
❗️ DO NOT MERGE
https://openscience.atlassian.net/browse/OSF-8685

# Purpose
~This is an exploratory ticket for examining using django-waffle with our Ember apps. Strategy is to create three API endpoints, one for samples, switches, and flags.  All types have an active field which shows whether that sample/switch/flag is active for the logged in user.  (Switches are active for everyone).  The ember app should have default values FLAG_DEFAULT, SAMPLE_DEFAULT, and SWITCH_DEFAULT for when the flag value in question is not found.
A route should try to load the sample/switch/flag in question. If it loads, use the active value as a conditional for whether or not to show certain pieces in the UI. If it's not found, use the default value listed in the config.~

Will be one endpoint that can be filtered against

# Summary of changes
- Added models/adapters/serializers for flag and switch. Sample not included, but would be similar
- Demonstrate disabling pagination on the nodes_list page, depending on whether the `pagination_flag` is active for the logged in user.

# Testing notes


pattisdr:
Was just an exploratory PR, when there's a front-end use-case for django-waffle, the plan is to have one API _waffle endpoint that can be filtered against to get switches, samples, and flags.

PR #285, opened 2017-10-16
[EOSF-116] Add institutions as a service in the navbar

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-763
https://openscience.atlassian.net/browse/EOSF-872

# Purpose
Add support for Institutions in the navbar, special casing it as it is not actually a separate service.

Note: Styling will look wrong until this is merged: https://github.com/CenterForOpenScience/osf-style/pull/106


baylee-d:
Changes look good!  But can you please resolve conflicts?

addon/components/new-osf-navbar/component.js
```diff
@@ -34,14 +34,20 @@ export default Ember.Component.extend(hostAppName, AnalyticsMixin, {
         if (appName === 'Dummy App') {
             appName = 'Home';
         }
+        if (appName === 'Ember OSF Web') {
+            if (window.location.pathname.indexOf('institutions') !== -1) {
+                appName = 'Institutions';
+            }
```
hmoco: Note to self, add:
```
} else {
    appName = 'Home'
}
```

addon/components/new-osf-navbar/component.js
```diff
@@ -34,6 +34,9 @@ export default Ember.Component.extend(hostAppName, AnalyticsMixin, {
         if (appName === 'Dummy App') {
             appName = 'Home';
         }
+        if (appName === 'Ember OSF Web') {
+            appName = window.location.pathname.indexOf('institutions') !== -1 ? 'Institutions' : 'Home';
```
brianjgeiger: Since we'll want this to go everywhere at once, we should probably make a waffle flag for the institutional dashboard. It doesn't have to necessarily keep the page from loading, but it should keep the navbar from having the navigational option.

addon/components/new-osf-navbar/component.js
```diff
@@ -34,6 +34,9 @@ export default Ember.Component.extend(hostAppName, AnalyticsMixin, {
         if (appName === 'Dummy App') {
             appName = 'Home';
         }
+        if (appName === 'Ember OSF Web') {
```
brianjgeiger: And this needs to be redone completely, yes? We have our own navbar, and this wouldn't be true in ember-osf anymore, I'm guessing.

brianjgeiger:
I think we're going to go with having you add in waffle handling for this (just the simple case needed here, not the full routing solution, of course). Please coordinate with @laurenbarker to determine which branch to target and if there are any extra requirements.

PR #287, opened 2017-10-18
Add old-file-browser to not break preprints.

hmoco:
# Purpose
Add the old `file-browser` as `old-file-browser` as breaking changes were made on it in the name of quickfiles, that turns out breaks preprints.

# Testing notes
Make a preprint from an existing node and an existing file. Make sure opening folders and selecting items works.

### DIFF
--- a/addon/components/old-file-browser/component.js
+++ b/addon/components/old-file-browser/component.js
@@ -1,0 +1,156 @@
+import Ember from 'ember';
+import layout from './template';
+
+import loadAll from 'ember-osf/utils/load-relationship';
+
+/**
+ * @module ember-osf
+ * @submodule components
+ */
+
+/*
+ * Wrapper for file items. Includes state for the item's row.
+ *
+ */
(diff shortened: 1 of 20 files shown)


addon/components/old-file-browser/style.scss
```diff
@@ -0,0 +1,34 @@
+.file-browser {
```
jamescdavis: Can we change this to old-file-browser to ensure no collision with new file-browser?

addon/components/old-file-browser/component.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+    return item;
+}
+
+/**
+ * File browser widget
+ *
+ * Sample usage:
+ * ```handlebars
+ * {{file-browser
```
jamescdavis: old-file-browser

addon/components/old-file-browser/style.scss
hmoco: right!

PR #289, opened 2017-10-19
[EOSF-882] fix discover page for Registries

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-882

# Purpose

Allow Registries discover page to return results once again

# Summary of changes

Handle case where themeProvider is not set.

# Testing notes

Make sure this doesn't break Preprints

### DIFF
--- a/addon/components/discover-page/component.js
+++ b/addon/components/discover-page/component.js
@@ -502,7 +502,7 @@ export default Ember.Component.extend(Analytics, hostAppName, {
             }
         });
         // For PREPRINTS and REGISTRIES. If theme.isProvider, add provider(s) to query body
-        if (this.get('themeProvider.name') !== null) {
+        if (this.get('themeProvider')) {
             const themeProvider = this.get('themeProvider');
             // Regular preprint providers will have their search results restricted to the one provider.
             // If the provider has additionalProviders, all of these providers will be added to the "sources" SHARE query


addon/components/discover-page/component.js
```diff
@@ -502,7 +502,7 @@ export default Ember.Component.extend(Analytics, hostAppName, {
             }
         });
         // For PREPRINTS and REGISTRIES. If theme.isProvider, add provider(s) to query body
-        if (this.get('themeProvider.name') !== null) {
+        if (this.get('themeProvider')) {
```
hmoco: It looks like this logic needs to be refactored. a bit otherwise https://github.com/CenterForOpenScience/ember-osf/pull/289/files#diff-ae3ef0b00a49fadec81c8b9b765a6d78R512 will never get hit, unless I'm missing something

hmoco: instead you can move `filters.push` into both `if` statements

hmoco: or I guess remove this line completely, and rewrite L519:
```
 if (sources.length) {
    filters.push({
        terms: {
            sources: sources
        }
    });
}
```

hmoco: welp this is what I get for not reading. `themeProvider` != `theme.isProvider`. Ignore my last few comments.

PR #291, opened 2017-10-24
[EOSF-863] If auth route is not authenticated, attempt to authenticate before sending to CAS

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-863

# Purpose
Using Firefox on staging and prod, if going to a page that has `cas-authenticated-route` mixin (before going to any other page of that ember app), the `beforeModel` hook would find the session to not be authenticated (the ember app still doesnt have an authenticated session) as the request to api would not resolve in time, and transition to CAS, which (since the user is authenticated) transitions back to the page, where the problem would reoccur infinitely. 

# Summary of changes
The mixin now attempts to authenticate if it finds the session to not be authenticated, prior to assuming that there is no auth and rerouting to cas

# Testing notes
Going to `osf.io/preprints/submit` and logging in, or going to osf.io and logging in then going to `osf.io/preprints/submit` WITHOUT going to any other preprints page while logged in, using Firefox, would consistently trigger infinite loading. 

Also logging in at `staging2.osf.io` and then typing in `staging2.osf.io/me/quickfiles`, or going straight there and logging in when prompted will trigger the same problem.

### DIFF
--- a/addon/mixins/cas-authenticated-route.js
+++ b/addon/mixins/cas-authenticated-route.js
@@ -26,7 +26,8 @@ export default Ember.Mixin.create({
     routing: Ember.inject.service('-routing'),
 
     /**
-      Checks whether the session is authenticated, and if it is not, redirects to the login URL. (Sending back to this page after a successful transition)
+      Checks whether the session is authenticated, and if it is not, attempts to authenticate it, and if that fails,
+      redirects to the login URL. (Sending back to this page after a successful transition)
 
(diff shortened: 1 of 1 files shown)


hmoco:
@jamescdavis @sloria, this theoretically fixes the problem, but it would be super useful to get this into a staging environment (staging2 maybe?) to test it, since locally this is not a problem but is a consistent problem on all deployed environments.

PR #331, opened 2017-12-14
[IN-71][Preprints] Add citation picker to citation-widget

hmoco:
## Purpose
It would be nice to allow users to retrieve citations other than MLA, chicago and APA, just like projects do, specially since the API, and citeproc-py engine are already in place. This adds that section to the citation-widget currently used in preprints.

## Summary of Changes
Add a citation picker to the citation-widget, allowing users to cite their preprints in any of our supported citation styles

Due to the large amount of citation styles, pressing something usual, like `a` and then waiting will make quite a few requests. Doing that a few times will probably trigger some 429 responses.

## Ticket

https://openscience.atlassian.net/browse/EOSF-883

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ *(asynchronous)*


addon/components/citation-widget/template.hbs
```diff
@@ -1,8 +1,24 @@
-<div class="citation-widget">
-    <div class="f-w-xl">APA</div>
-    <span>{{apa}} </span>
-    <div class="f-w-xl m-t-md">MLA</div>
-    <span> {{mla}} </span>
-    <div class="f-w-xl m-t-md">Chicago</div>
-    <span> {{chicago}} </span>
+<div class="">
```
baylee-d: Does this need to have an empty class identifier?

addon/components/citation-widget/component.js
```diff
@@ -50,5 +53,25 @@ export default Ember.Component.extend({
                 .ajax(`${citationLink}${linkSuffix}/`, 'GET')
                 .then(resp => this.set(attr, resp.data.attributes.citation));
         }
-    }
+    },
+    _citationText: Ember.observer('selectedStyle', function() {
+        const citationLink = this.get('node.links.relationships.citation.links.related.href');
```
baylee-d: Since this is also being set up in `didReceiveAttrs()`, would it make more sense to make a new variable `citationLink: null` up above and change the value of it in `didReceiveAttrs()` if needed?  Then you'd only need to make a call to that variable here.

addon/components/citation-widget/component.js
```diff
@@ -43,12 +46,33 @@ export default Ember.Component.extend({
...
 
         for (const { linkSuffix, attr } of citationStyles) {
             this.get('store')
                 .adapterFor('node')
                 .ajax(`${citationLink}${linkSuffix}/`, 'GET')
                 .then(resp => this.set(attr, resp.data.attributes.citation));
         }
-    }
+    },
+    _citationText: Ember.observer('selectedStyle', function() {
```
baylee-d: Is there any way you can use a `computed` here instead of an `observer`?

app/styles/app.scss
```diff
@@ -0,0 +1,2 @@
+
+@import "ember-power-select";
```
baylee-d: I took back what I said earlier since there should be a fix for this now.  You shouldn't have to do the `@import "ember-power-select"` with the newest version of ember-osf.

package.json
```diff
@@ -84,6 +84,7 @@
     "ember-ace": "^1.2.0",
     "ember-bootstrap-datepicker": "^2.0.3",
     "ember-cli-babel": "6.4.1",
+    "ember-concurrency": "0.8.12",
```
baylee-d: Did you mean to have `ember-concurrency` twice here in the same section?

addon/components/citation-widget/template.hbs
```diff
@@ -1,8 +1,24 @@
...
-    <div class="f-w-xl">APA</div>
-    <span>{{apa}} </span>
-    <div class="f-w-xl m-t-md">MLA</div>
-    <span> {{mla}} </span>
-    <div class="f-w-xl m-t-md">Chicago</div>
-    <span> {{chicago}} </span>
+<div>
+    <div id="citationList" class="m-b-md">
+        <div class="f-w-xl">APA</div>
+        <span>{{apa}} </span>
```
baylee-d: Should maybe add a space in front of the APA citation to keep it consistent with the other styles

package.json
hmoco: nope! fixed.

PR #338, opened 2018-01-11
[IN-77][Ember-OSF] Add 'lazy-options' as a ember-power-select plugin for lazy loading.

adlius:
## Purpose

For [EOSF-712](https://openscience.atlassian.net/browse/EOSF-712), the purpose is to implement lazy loading when loading user nodes for selection. This is the first of two PRs that implement a `lazy-options` plugin to override `optionsComponent` of ember-power-select, which does not support lazy loading natively.

## Summary of Changes

1. Add a `jquery-checkinview.js` as a jquery extension for checking whether a certain element in in view.
    - Add `jquery-extenstion.js` to initialize t he extension.

2. Add `lazy-options` component.
    - To use this component, user need to supply two boolean parameters `canLoadMore`, `isLoading` and an action `loadMore`.
    - As seen in `lazy-options/template.hbs`, an element with css class`ember-power-select-option--load-more` will be appended to the end of the options list if `canLoadMore` is set.
    - As seen in `lazy-options/component.js`, if element with css class `ember-power-select-option--load-more` is found and in view (meaning that user has scrolled to the bottom of the current options list), `loadMore` action will be triggered.
    - To use this plugin for ember-power-select, override the `optionsComponent` with this plugin. (See the second PR for examples.)

3. Add a `loadPage` function to `load-relationship.js` to load specific page of related model.
    - This is needed for lazy loading user nodes in action `loadMore`, which should get the next page of API results and determine whether more results are available.

4. Add `ember-concurrency` as a dependency.

## Ticket

https://openscience.atlassian.net/browse/IN-77

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


addon/components/lazy-options/template.hbs
```diff
@@ -0,0 +1,43 @@
...
+            aria-selected="{{ember-power-select-is-selected opt select.selected}}"
+            aria-disabled={{ember-power-select-true-string-if-present opt.disabled}}
+                aria-current="{{eq opt select.highlighted}}"
+            data-option-index="{{groupIndex}}{{index}}"
+            role="option">
+            {{yield opt select}}
+        </li>
+    {{/if}}
+{{/each}}
+{{#if (or isLoading)}}
```
baylee-d: It looks like there's only one condition in this if.  Was there a reason to have an `or`?

addon/components/lazy-options/template.hbs
```diff
@@ -0,0 +1,43 @@
+<!--  https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/options.hbs -->
+{{#if select.loading}}
```
baylee-d: A couple of things/questions:

1. Did you mean to have two different loading messages here? 
2. Is there a difference between `select.loading` and `isLoading`?  If they're both the same, it might make more sense to use one or the other, or just use `isLoading` as an alias of `select.loading`.

addon/instance-initializers/jquery-extensions.js
```diff
@@ -0,0 +1,11 @@
+import { checkInView } from '../utils/jquery-checkinview';
+
+export function initialize() {
+    // Add jQuery extension calls here
+    checkInView();
+}
+
+export default {
+    name:       'jq-extensions',
```
baylee-d: Should probably remove the indentation here

addon/components/lazy-options/template.hbs
adlius: It seems that I removed one of the variables and forgot to get rid of the 'or'...

addon/components/lazy-options/template.hbs
adlius: This template is mostly the same as the options template for ember-power-select (https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/options.hbs). `select.loading` is used by ember-power-select to know whether the options are loading. `isLoading` is for whether additional options is currently loading.

baylee-d: In that case, would it work to make a `computed.or` in `lazy-options/component.js` that looked at `isLoading` and `select.loading`?  Then you can just use that property to determine if anything is loading at all.  

That or you could maybe only have one statement in the template that looked at `{{#if (or select.loading isLoading)}}`

addon/utils/load-relationship.js
```diff
@@ -16,6 +16,29 @@ export function loadRelation(model, relationship) {
     return ArrayPromiseProxy.create({promise});
 }
 
+export function loadPage(model, relationship, pageSize, page, options = {}) {
+    var query = {
```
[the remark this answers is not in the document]
adlius: variables declared by `var` are global to entire function while those declared by `let` are limited to the block scope. You're right, there is no need to use `var` here. I was copy and pasting from `loadAll`, which does need the global scope for `query`.

addon/components/lazy-options/template.hbs
adlius: The idea is that they are indeed two different loading messages for different circumstances.

alexschiller:
@adlius Travis failing lint

alexschiller:
@adlius  Switching the branch to point to next-interfaces, which will redo travis. Here was the error:

$ eslint .
/home/travis/build/CenterForOpenScience/ember-osf/addon/utils/jquery-checkinview.js
  16:15  error  'contBottom' is assigned a value but never used  no-unused-vars

alexschiller:
Fixed changelog location, merging.

PR #340, opened 2018-01-11
[MOD-176][Feature] Replace has-many-query with queryHasMany

aaxelb:
## Purpose
Move the [`queryHasMany` helper from ember-osf-reviews](https://github.com/CenterForOpenScience/ember-osf-reviews/blob/develop/app/services/store.js#L21) to ember-osf. Provides a way to query relationships without going through ember-data, which sometimes decides it has all the data it needs and doesn't make the request.

## Summary of Changes
- Add `queryHasMany` method to `osf-model`
  - Allows querying relationships with custom query params and ajax options
    e.g. `user.queryHasMany('nodes', { filter: { title: "My title!", public: true } })`
- Remove ember-data-has-many-query

Code that uses `model.query()` has to be updated to use `model.queryHasMany()`. In most cases, it's a drop-in replacement, but I didn't want to replace it silently in case someone tries to query a `belongsTo` or somehow depends on `ember-data-has-many-query`-specific behavior. 

## Ticket

https://openscience.atlassian.net/browse/MOD-176

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY
- [x] testable and includes test(s)
- [x] changes described in `CHANGELOG.md`

### DIFF
--- a/addon/models/osf-model.js
+++ b/addon/models/osf-model.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import DS from 'ember-data';
-import { OSFAdapter } from 'ember-osf/adapters/osf-adapter';
-import { authenticatedAJAX } from 'ember-osf/utils/ajax-helpers';
+import OSFAdapter from '../adapters/osf-adapter';
+import { authenticatedAJAX } from '../utils/ajax-helpers';
 
 /**
  * @module ember-osf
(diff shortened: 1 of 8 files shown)


addon/models/osf-model.js
```diff
@@ -65,11 +65,55 @@ export default DS.Model.extend(HasManyQuery.ModelMixin, {
...
+                Ember.run.bind(this, this.__queryHasManyDone, resolve),
+                reject
+            );
+        });
+
+        const ArrayPromiseProxy = Ember.ArrayProxy.extend(Ember.PromiseProxyMixin);
+        return ArrayPromiseProxy.create({ promise });
+    },
+
+    __queryHasManyDone(resolve, payload) {
```
jamescdavis: I like the double underscore prefix for "superprivate" methods that should never be called directly (only as a callback). AFAIK, this is not a convention in JS, but I like it.

PR #341, opened 2018-01-19
[EMB-60] Prevent user from selecting multiple files when clicking upload button.

adlius:
## Purpose

Currently for quickfiles, uploading multiple files by drag'n'drop is not allowed and would show "Cannot upload multiple files". However, this is not the case when user clicks the "upload" button and select multiple files to upload. This PR fix this problem by not allowing users to select multiple files for upload.

## Summary of Changes

Dropzone.js handles drag'n'drop upload differently than clicking the "upload" button for upload. Therefore, there is not an event handler that we can override to prevent multiple uploads through clicking the "upload" button. However, we could remove the "multiple" attribute from the hidden `<input>` element Dropzone uses, thus preventing users from selecting multiple files. And that is what is done in this PR.

None.

## Ticket

https://openscience.atlassian.net/browse/EMB-60

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/components/dropzone-widget/component.js
+++ b/addon/components/dropzone-widget/component.js
@@ -80,6 +80,14 @@ export default Ember.Component.extend({
             }
         });
 
+        // Dropzone.js does not have an option for disabling selecting multiple files when clicking the "upload" button.
+        // Therefore, we remove the "multiple" attribute for the hidden file input element, so that users cannot select
+        // multiple files for upload in the first place.
+        let clickableElement = this.get('clickable');
+        if (this.options.preventMultipleFiles && clickableElement){
+            Ember.$('.dz-hidden-input').removeAttr('multiple');
+        }
+
         this.set('dropzoneElement', drop);
 
         // Set osf session header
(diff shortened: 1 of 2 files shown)


addon/components/dropzone-widget/component.js
```diff
@@ -80,6 +80,14 @@ export default Ember.Component.extend({
             }
         });
 
+        // Dropzone.js does not have an option for disabling selecting multiple files when clicking the "upload" button.
+        // Therefore, we remove the "multiple" attribute for the hidden file input element, so that users cannot select
+        // multiple files for upload in the first place.
+        let clickableElement = this.get('clickable');
+        if (this.options && this.options.preventMultipleFiles && clickableElement){
```
jamescdavis: Should use this.get() e.g. `this.get('options.preventMultipleFiles')`

addon/components/dropzone-widget/component.js
```diff
@@ -80,6 +80,14 @@ export default Ember.Component.extend({
             }
         });
 
+        // Dropzone.js does not have an option for disabling selecting multiple files when clicking the "upload" button.
+        // Therefore, we remove the "multiple" attribute for the hidden file input element, so that users cannot select
+        // multiple files for upload in the first place.
+        let clickableElement = this.get('clickable');
```
jamescdavis: This variable is only used once and is probably not necessary.

addon/components/dropzone-widget/component.js
jamescdavis: `if (this.get('options.preventMultipleFiles') && this.get('clickableElement')) {`

jamescdavis:
Passing build: https://travis-ci.org/adlius/ember-osf/builds/332877360

PR #345, opened 2018-01-31
[IN-78][Emberosf] Fix focusOut issue

adlius:
## Purpose

Currently, when the user manually edits the "Original publication date" field on preprint submission/edit page, the value of the property does not change when the element loses focus. This PR fixes the problem.

## Summary of Changes

A `focusOut` event handler is added to parse and set the value from manual input.

For an existing preprint, make sure that when you manually change date value in the "Original publication date" field (not using datepicker) and then click "save", the change is reflected in the preprint detail page.

## Ticket

https://openscience.atlassian.net/browse/IN-78

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/components/validated-input/component.js
+++ b/addon/components/validated-input/component.js
@@ -29,6 +29,16 @@ export default Ember.Component.extend({
         defineProperty(this, 'value', computed.alias(`model.${valuePath}`));
     },
 
+    // This is a to make sure we parse and set the date value from manual inputs.
+    focusOut: function () {
+        if (this.get('isDateField')) {
+            let date = Date.parse(this.$('input')[0].value);
+            if (!isNaN(date)) {
+                this.set('value', new Date(date));
+            }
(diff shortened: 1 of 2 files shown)


addon/components/validated-input/component.js
```diff
@@ -29,6 +29,16 @@ export default Ember.Component.extend({
...
 
+    // This is a to make sure we parse and set the date value from manual inputs.
+    focusOut: function () {
+        if (this.get('isDateField')) {
+            let date = Date.parse(this.$('input')[0].value);
+            if (!isNaN(date)) {
+                this.set('value', new Date(date));
+            }
+        }
+    },
```
jamescdavis: This validation should be done with [ember-cp-validations](https://github.com/offirgolan/ember-cp-validations) instead

PR #348, opened 2018-02-02
[EMB-17] updates for Quick Files move to project

jamescdavis:
## Purpose

Requested changes for Quick Files move to project workflow

## Summary of Changes

* verbiage changes (see ticket)
* change list of projects to include all projects/components for which the user has **write** access
* when moving to a new component, set the component category to `uncategorized` instead of `project`

`node.addChild(_, _, null)` no longer sets the new component's category to 'project' (will set to `null,` which is invalid). Not passing in the category at all (as done in Preprints) will still default to 'project'.

Double check verbiage display logic makes sense.

## Ticket

https://openscience.atlassian.net/browse/EMB-17

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


addon/locales/en/translations.js
```diff
@@ -223,18 +223,24 @@ export default {
...
                 noProjectsExistError: 'You have no available projects. Go back to create a new project.',
                 organizeProject: 'Organize',
-                organizeParentNode: 'You can organize your file by storing it in the project or in its own component. If you select "Make a new component", the file will be stored in a new component inside the project. If you select "Use the current component", the file will be stored inside the project. If you are unsure, select "Make a new component".',
-                organizeChildNode: 'You can organize your file by storing it in this component or in its own component. If you select "Make a new component", the file will be stored in a new component inside this component. If you select "Use the current component", the file will be stored inside this component. If you are unusure, select "Make a new component".',
+                organizeParentNode: 'You can organize your file by storing it in the project or its own component.',
+                organizeChildNode: 'You can organize your file by storing it in the component or its own component.',
```
jamescdavis: Hmm, not sure about s/this/the/ here. The requested changes just said to basically s/project/component/, but maybe this > the here?

addon/models/node.js
```diff
@@ -167,11 +167,11 @@ export default OsfModel.extend(FileItemMixin, {
         return promise;
     },
 
-    addChild(title, description, category, isPublic) {
+    addChild(title, description=null, category='project', isPublic=false) {
```
jamescdavis: Hmm, just realized that having `isPublic=false` as a default will probably not do the right thing for preprints.

jamescdavis:
Passing build: https://travis-ci.org/jamescdavis/ember-osf/builds/336723278
(but more changes might be coming)

addon/components/file-browser/component.js
```diff
@@ -472,7 +472,7 @@ export default Ember.Component.extend(Analytics, {
     },
     _createComponent(title) {
         return this.get('node')
-            .addChild(title, null, null, true)
+            .addChild(title, null, '', true)
```
aaxelb: I don't understand, why do we need need/want `category: ''`? According to [API docs](https://developer.osf.io/#operation/nodes_create) that's invalid, but it sounds like it actually sets `category: 'uncategorized'`? If that's right, can we set that explicitly instead?

aaxelb: Oh, never mind, [I see](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/settings/defaults.py#L255) ([twice?](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L213)).

addon/models/node.js
aaxelb: Yeah, changing interfaces is worrisome. Might be better to replace `category || 'project'` below with something like `category || category === '' ? category : 'project'`. Though that might be getting gross enough it's worth upgrading to 
```
if (!category && category !== '') {
    category = 'project';
}

Comment (aaxelb) [reply addon/models/node.js]: (or I guess `typeof(category) === 'undefined' ? 'project' : category`)

File: addon/components/file-browser/component.js (hunk 3 above)
Comment (jamescdavis) [reply addon/components/file-browser/component.js]: Nope, ['' == 'Uncategorized'](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/node.py#L213)

Comment (jamescdavis) [reply addon/components/file-browser/component.js]: screen shot 2018-02-02 at 4 38 01 pm

File: addon/models/node.js [hunk 4]
```diff
@@ -167,11 +167,11 @@ export default OsfModel.extend(FileItemMixin, {
         return promise;
     },
 
-    addChild(title, description, category, isPublic) {
+    addChild(title, description=null, category='project', isPublic) {
```
jamescdavis: I went ahead and removed the default value for isPublic so as to not change the existing behavior and potentially break things. It is always specified when called in Quick Files move to project anyways.

addon/locales/en/translations.js
aaxelb: I agree, this language is confusing.

addon/locales/en/translations.js
```diff
@@ -223,18 +223,24 @@ export default {
...
                 createComponent: 'Create component',
                 couldNotCreateComponent: 'Could not create component. Please try again.',
                 couldNotCreateProject: 'Could not create project. Please try again.',
                 useCurrentComponent: 'Use current component',
                 useCurrentProject: 'Use current project',
-                convertOrCopyMessage: 'You have selected and organized your file. Clicking "Move file" will immediately make changes to your OSF project and move your file. You can continue viewing and editing your file from your project.',
-                noLongerPublicWarning: 'Files moved to private projects/components will no longer be public or discoverable by others.',
+                convertOrCopyMessage: {
+                    project: 'Clicking "Move file" will immediately make changes to your OSF project and move your file.',
+                    component: 'Clicking "Move file" will immediately make changes to your OSF component and move your file.',
```
aaxelb: Minor language question: Will moving a file to a project/component actually change that node in any way? This was true for preprints but doesn't seem true for quickfiles.

addon/locales/en/translations.js
```diff
@@ -223,18 +223,24 @@ export default {
...
+                },
+                noLongerPublicWarning: {
+                    project: 'Files moved to private projects will no longer be public or discoverable by others.',
+                    component: 'Files moved to private components will no longer be public or discoverable by others.'
```
aaxelb: Are quickfiles discoverable/searchable to others? This seems like another thing that was true for preprints but isn't true for quickfiles.

addon/models/node.js
aaxelb: Works for me 👍

addon/locales/en/translations.js
brianjgeiger: They are, because you can either use the API or go to the user's profile page to see their quickfiles.

aaxelb: Ah okay, that makes sense.

addon/components/project-selector/template.hbs
```diff
@@ -85,13 +85,21 @@
...
-            </p>
-            <p class="text-danger m-t-sm">
-                {{#unless isProjectPublic}}
-                    {{t "eosf.components.moveToProject.noLongerPublicWarning"}}
-                {{/unless}}
+                {{#if isChildNode}}
+                    {{t "eosf.components.moveToProject.convertOrCopyMessage.component"}}
+                {{else}}
+                    {{t "eosf.components.moveToProject.convertOrCopyMessage.project"}}
+                {{/if}}
```
jamescdavis: One of these will always display once the user has selected where to move the file, either to a new component under an existing node (`willCreateComponent`) or to an existing node (`willMoveToNode`)

addon/components/project-selector/template.hbs
```diff
@@ -85,13 +85,21 @@
...
             </p>
+            {{#unless isProjectPublic}}
+                <p class="text-danger m-t-sm">
+                    {{#if isChildNode}}
+                        {{t "eosf.components.moveToProject.noLongerPublicWarning.component"}}
+                    {{else}}
+                        {{t "eosf.components.moveToProject.noLongerPublicWarning.project"}}
+                    {{/if}}
```
jamescdavis: One of these will always display once the user has selected where to move the file and the destination is (or will be) private.

addon/locales/en/translations.js
jamescdavis: I guess technically it won't change the project/component, but does change its relationships?

addon/locales/en/translations.js
jamescdavis: Yes, it is, but leaving as-is for now.

addon/locales/en/translations.js
brianjgeiger: It will create a log on the node, thus updating its date_modified. So from a user's perspective, it makes a change.

PR #350, opened 2018-02-07
[EMB-105] Add isUploading computed and use to disable upload button

jamescdavis:
## Purpose

Disable upload button while a file is uploading.

## Summary of Changes

* add isUploading computed based on whether the uploading list is not empty
* use it to disable the upload button

Make sure you can't click the upload button while an upload is in progress.

## Ticket

https://openscience.atlassian.net/browse/EMB-105

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/components/file-browser/component.js
+++ b/addon/components/file-browser/component.js
@@ -120,6 +120,7 @@ export default Ember.Component.extend(Analytics, {
     sortingBy: 'itemName',
     sortingOrder: 'asc',
     uploading: Ember.A(),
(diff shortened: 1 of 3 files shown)


aaxelb:
`file-browser` overrides some of bootstrap's disabled button styles, so the button doesn't look/act disabled except that it's half a shade lighter and clicking it does nothing:
disabled enabled
It should have `cursor: not-allowed;` (or `default`) and probably non-green text?

jamescdavis:
Better?

PR #356, opened 2018-02-13
[IN-103][Ember-OSF] Use currentUserId for ElasticSearch preference

fabmiz:
## Purpose

Fix current issue where share elasticSearch returns inconsistent
result ordering.

See here for more details
https://www.elastic.co/guide/en/elasticsearch/reference/current/consistent-scoring.html#_scores_are_not_reproducible

## Summary of Changes
See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-preference.html for suggested solution.

Modify searchUrl to use currentUserId as preference custom string in case user
is logged in. Otherwise, use a random string.

This ticket affects the preprints discover/search page.
Test this within the developer console, networks tab:
- Clear the network tab
- Upon searching a term, check the elasticSearch requests to see if the preference
URL parameter was added and looks something like this (in chrome console)
i) logged-in user:
`_search?preference=%24%5Eezcos%23%40`, `ezcos` being the current user id.
ii) Not logged-in user:
_search?preference="some random string"

- In each case, try the same search several times (hitting enter in search bar) in a row and make sure the same preference string is sent.

While the goal is to get the same ordering of search results for the same search query, it might be very hard to compare before and after so we can just leave it up to elasticSearch to do what its thing.

## Ticket

[IN-103](https://openscience.atlassian.net/browse/IN-103)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added "choose your custom citation" section to citation-widget
 - `facebookAppId` field to `preprint-provider` model
+- Add ElasticSearch preference (ES) key to preprints ES search requests for reproducible results ordering.
 
 ## [0.14.0] - 2018-02-07
 ### Added
@@ -24,7 +25,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Move function for `file` model
 - `ember-collapsable-panel` and `ember-power-select` packages
 - Styling on `project-selector` buttons that are selected
-- Add ElasticSearch preference (ES) key to preprints ES search requests for reproducible results ordering.
 
 ### Changed
 - Node `addChild` model function to create public child elements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Move function for `file` model
 - `ember-collapsable-panel` and `ember-power-select` packages
 - Styling on `project-selector` buttons that are selected
+- Add ElasticSearch preference (ES) key to preprints ES search requests for reproducible results ordering.
 
 ### Changed
 - Node `addChild` model function to create public child elements
--- a/addon/services/current-user.js
+++ b/addon/services/current-user.js
@@ -65,17 +65,38 @@ export default Ember.Service.extend({
     }),
 
     /**
-     * Custom string appended as URL param to ElasticSearch requests for reproducible search results ordering.
-     * If user is logged in, use their userID, otherwise return generated string.
+     * Return a simple hash of the currentUser ID if user is logged in, otherwise return a generated random string.
+     * sessionKey can be used to identify the current session or any general purposes.
+     * For Elasticsearch requests, sessionKey is used as the "preference" URL parameter to ensure reproducible search results ordering.
      *
-     * @property esPreferenceKey
+     * @property sessionKey
      * @return {String}
      */
-    esPreferenceKey: Ember.computed('currentUserId', function() {
+    sessionKey: Ember.computed('currentUserId', function() {
         let currentUserId = this.get('currentUserId');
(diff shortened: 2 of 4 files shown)


addon/services/current-user.js
```diff
@@ -63,4 +63,16 @@ export default Ember.Service.extend({
...
     }),
+
+    /**
+     * @property esPreferenceKey
+     * @return {String}
+     */
+    esPreferenceKey: Ember.computed('currentUserId', function() {
+        let currentUserId = this.get('currentUserId');
+        if (currentUserId) {
+            return encodeURIComponent(`$^${currentUserId}#@`);
```
laurenbarker: What is `$^...#@` doing here?

addon/services/current-user.js
```diff
@@ -63,4 +63,16 @@ export default Ember.Service.extend({
             promise: this.load().catch(() => null),
         });
     }),
+
+    /**
+     * @property esPreferenceKey
+     * @return {String}
+     */
+    esPreferenceKey: Ember.computed('currentUserId', function() {
```
laurenbarker: A comment might be good here explaining why we need this.

laurenbarker:
Oh also this will need some QA notes (forgot to mention that)

addon/services/current-user.js
aaxelb: Just a drive-by thought, for what it's worth:

It's generally nicer to name something after what it *is* or what it *does*, not how/when it's used. This is a user- or session-specific key that is currently used only for ES preference, but could be used for other things in the future.

Something like `userOrSessionKey` (or something better, I don't know, naming is hard) would be less likely to cause confusion/annoyance/work in the future. Then, as @laurenbarker said, a comment would be a good place to explain how we expect to use it, why it exists.

addon/services/current-user.js
fabmiz: we'd decided to use a hash of the currentUserId but I could not find a simple (native) way to do this without adding another ember addon. `encodeURIComponent` only encode certain chars such as those I added. the outcome is like so (looks kinda hashed maybe...)
`> encodeURIComponent('$^ezcos#@')
'%24%5Eezcos%23%40'`. We can definitely just return the `currentUserId` if you think it's overkill.

alexschiller: If we really think a hash is necessary there are some pretty simple options. Though I'd just add a helper function instead of adding a string method. 
https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery

If you want a hash do something like that. I don't think we *need* to obfuscate the guid, but I could see reasons why it would be nice. It is a pretty simple change, so just do it, having that kind of session id could be desirable for other things as well.

addon/services/current-user.js
```diff
@@ -63,4 +63,19 @@ export default Ember.Service.extend({
...
     }),
+
+    /**
+     * Custom string appended as URL param to ElasticSearch requests for reproducible search results ordering.
+     * If user is logged in, use their userID, otherwise return generated string.
+     *
+     * @property esPreferenceKey
+     * @return {String}
+     */
+    esPreferenceKey: Ember.computed('currentUserId', function() {
```
alexschiller: I agree with Abram's comment that this can be named more generically. Especially if you use a more proper hash I think it is safe to call this `sessionKey` and adjust the comment to say it is just a simple hash of the username that can identify the current session and that it is used for "... reproducible search results ordering..." etc.

alexschiller:
@fabmiz conflicts when I switched branch to next-interfaces

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Move function for `file` model
 - `ember-collapsable-panel` and `ember-power-select` packages
 - Styling on `project-selector` buttons that are selected
+- Add ElasticSearch preference (ES) key to preprints ES search requests for reproducible results ordering.
```
alexschiller: This need to be moved up to unreleased 0.15.0

PR #370, opened 2018-03-14
[IN-6][Ember-OSF] Add model for moderator

laurenbarker:
## Purpose
Allow preprint providers to add and remove moderators using the moderation app.

## Summary of Changes
* Add moderator model, serializer, and adapter.

Should be covered by testing the Reviews side.

## Ticket

https://openscience.atlassian.net/browse/IN-6

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/adapters/moderator.js
+++ b/addon/adapters/moderator.js
@@ -12,15 +12,15 @@ export default OsfAdapter.extend({
         } else if (requestType === 'createRecord' && snapshot.record.provider) {
             provider = snapshot.record.provider;
             delete snapshot.record.provider;
-            return this._buildURL(modelName) + `/${provider}/moderators/`;
+            return `${this._buildURL(modelName)}/${provider}/moderators/`;
         } else if (snapshot.adapterOptions && snapshot.adapterOptions.provider) {
             provider = snapshot.adapterOptions.provider;
             delete snapshot.adapterOptions.provider;
             if (requestType === 'updateRecord' || requestType === 'deleteRecord' || requestType === 'findRecord') {
-                return this._buildURL(modelName) + `/${provider}/moderators/${id}/`;
+                return `${this._buildURL(modelName)}/${provider}/moderators/${id}/`;
             }
         }
 
-        return this._super(...arguments) + `${provider}/moderators/`;
+        return `${this._super(...arguments)}${provider}/moderators/`;
     }
 });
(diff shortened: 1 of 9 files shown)


addon/adapters/moderator.js
```diff
@@ -0,0 +1,26 @@
+import OsfAdapter from './osf-adapter';
+
+export default OsfAdapter.extend({
+    pathForType() {
+        return 'preprint_providers';
+    },
+    buildURL(modelName, id, snapshot, requestType, query) {
```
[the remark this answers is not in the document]
laurenbarker: It might also be more difficult to maintain. Since this is endpoint is pretty hacky I would hope that it is temporary.

addon/adapters/moderator.js
```diff
@@ -0,0 +1,26 @@
...
+        let provider = '';
+        if (query) {
+            provider = query.provider;
+            delete query.provider;
+        } else if (requestType === 'createRecord' && snapshot.record.provider) {
+            provider = snapshot.record.provider;
+            delete snapshot.record.provider;
+            return this._buildURL(modelName) + `/${provider}/moderators/`;
```
alexschiller: Minor; would this be better as one template literal instead of `+` and a template literal?
```javascript
return `${this._buildURL(modelName)}/${provider}/moderators/`;
```

addon/adapters/moderator.js
```diff
@@ -0,0 +1,26 @@
+import OsfAdapter from './osf-adapter';
+
+export default OsfAdapter.extend({
+    pathForType() {
+        return 'preprint_providers';
+    },
+    buildURL(modelName, id, snapshot, requestType, query) {
+        let provider = '';
+        if (query) {
+            provider = query.provider;
```
alexschiller: A few of these could be destructured assignment I think
```javascript
{ provider } = query;
```

addon/adapters/moderator.js
```diff
@@ -0,0 +1,26 @@
...
+            delete query.provider;
+        } else if (requestType === 'createRecord' && snapshot.record.provider) {
+            provider = snapshot.record.provider;
```
alexschiller: { provider } = snapshot.record;

addon/adapters/moderator.js
laurenbarker: Destructuring didn't seem to work :/

`{ provider } = query;` caused a syntax error and `let { provider } = query;` resulted in provider not being set. Did I miss something?

PR #402, opened 2018-07-14
[IN-102][EOSF] Add search-users model 

alexschiller:
## Purpose

Need way to hit the search users query endpoint

Added model/tests for search-users, which extend users.

## Testing Notes

None here, testing notes on preprints side ticket

## Ticket

https://openscience.atlassian.net/browse/IN-102

## Notes for Reviewer

Is this the right way to do this? Mostly pulled from other models so there may be some cargo culty copy pasta. Is there a better way to say another endpoint uses the user model without extending an empty model? Basically, did I do this right?

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


alexschiller:
Closing in favor of #416 since this was rewritten

PR #420, opened 2018-10-26
[EMB-400] Change the endpoint for the authenticator from `/v2/users/me/` to `/v2/`

adlius:
## Purpose

Currently, `ember-osf` authenticator asks `/v2/users/me/` endpoint for current user's information. If the user is not authenticated, the endpoint returns a 401, which is not good.
Therefore, we change to using `/v2/` which would not return a 401 even if the user is not authenticated.

In `osf-cookie.js`, the endpoint of the AJAX request is changed to `/v2/`.

## Ticket

https://openscience.atlassian.net/browse/EMB-400

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/authenticators/osf-cookie.js
+++ b/addon/authenticators/osf-cookie.js
@@ -36,7 +36,11 @@ export default Base.extend({
             // Push the result into the store for later use by the current-user service
             // Note: we have to deepcopy res because pushPayload mutates our data
             // and causes an infinite loop because reasons
-            this.get('store').pushPayload(Ember.copy(res['meta']['current_user'], true));
+            if (res['meta']['current_user'] !== null ){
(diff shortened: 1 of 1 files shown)


addon/authenticators/osf-cookie.js
```diff
@@ -36,7 +36,7 @@ export default Base.extend({
             // Push the result into the store for later use by the current-user service
             // Note: we have to deepcopy res because pushPayload mutates our data
             // and causes an infinite loop because reasons
-            this.get('store').pushPayload(Ember.copy(res, true));
+            this.get('store').pushPayload(Ember.copy(res['meta']['current_user'], true));
```
aaxelb: If they're not logged in, `res.meta.current_user` will be missing/null. Might be better not to `pushPayload` in that case

PR #428, opened 2019-02-21
[EMB-531][EOSF] API unavailable message fix

adlius:
## Purpose
In the previous implementation of this ticket, it seems that the `error` event handler in child routes will conflict will preempt the route redirection in the `catch` block of the `beforeModel` hook. Weird thing is that it only happens in a prod build but not a development build. This PR attempt to fix the problem by utilizing the `error` event handler in the route. See more below.

Instead of catching the error after returning the promise in the `beforeModel` hook and transition to a `error-no-api` route, we move the transition to the `error` handler. There are two benefits to this:
- `error` handler in the route catches error occurred in the lifecycle hooks, so it's equivalent to the catch block.
- It also catches errors bubbled up from child route's

## Ticket

https://openscience.atlassian.net/browse/EMB-531

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/addon/mixins/osf-cookie-login-route.js
+++ b/addon/mixins/osf-cookie-login-route.js
@@ -25,15 +25,13 @@ export default Ember.Mixin.create({
         if (this.get('session.isAuthenticated')) return;
 
         // Block transition until auth attempt resolves. If auth fails, let the page load normally.
-        return this.get('session').authenticate('authenticator:osf-cookie')
-            .catch(err => {
-                Ember.Logger.log('Authentication failed: ', err);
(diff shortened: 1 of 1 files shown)


jamescdavis:
This looks good, but we need to make sure it's also compatible with Reviews. But also, Reviews just currently white screens when the API server is blocked, so just need to make sure not-found still works properly and this doesn't break anything.

adlius:
@jamescdavis It should work with reviews with no problem. I've added a PR (https://github.com/CenterForOpenScience/ember-osf-reviews/pull/161) to fix a previously existing bug where I used `replaceWith` in the controller.

PR #458, opened 2022-01-14
[ENG-23] Serialize meta for preprint findRecord

futa-ikeda:
## Purpose
- Allow a preprint's metrics metadata to be serialized in the `apiMeta` property

- Normalize `findRecord` responses to include the response.meta into an `apiMeta` attribute so it can be used later in the app (specifically preprint's views/download metrics)

## Testing Notes
- this change is additive and doesn't change any existing user-facing behavior

## Ticket

https://openscience.atlassian.net/browse/ENG-23

## Notes for Reviewer
- I've opted to store the response.meta object into the `apiMeta` attribute as this seemed like the easiest way to access the impact metrics for a given preprint. I looked at using store.findMetadataFor('preprint'), but it seems this is deprecated.

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
         key: cached_node_modules_${{ hashFiles('**/yarn.lock') }}
         restore-keys: cached_node_modules_
     - if: steps.cache-node-modules.outputs.cache-hit != 'true'
-      run: yarn --prefer-offline --ignore-engines
+      run: yarn --prefer-offline --frozen-lockfile --ignore-engines
 
   test-with-coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
         key: cached_node_modules_${{ hashFiles('**/yarn.lock') }}
         restore-keys: cached_node_modules_
     - if: steps.cache-node-modules.outputs.cache-hit != 'true'
-      run: yarn --prefer-offline --frozen-lockfile --ignore-engines
+      run: yarn --prefer-offline --ignore-engines
 
   test-with-coverage:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         key: cached_node_modules_${{ hashFiles('**/yarn.lock') }}
         restore-keys: cached_node_modules_
     - if: steps.cache-node-modules.outputs.cache-hit != 'true'
-      run: yarn --prefer-offline --frozen-lockfile --ignore-engines
+      run: yarn --prefer-offline --ignore-engines
```
brianjgeiger, line 29: Don't we prefer frozen lockfiles? What prompted the change?

brianjgeiger: But you added a new lock file to this PR. Shouldn't that have all the updates it needs?

futa-ikeda: Ahh, I didn't push the updated `yarn.lock` after moving faker... Will rectify

futa-ikeda: realized that a lot of the yarn.lock file changes came from when I updated ember-faker, which I didn't need to do, so force-pushed without that commit in there to keep things a bit easier to read

brianjgeiger (dismissed):

brianjgeiger (approved):

addon/serializers/osf-serializer.js
```diff
@@ -141,5 +141,13 @@ export default DS.JSONAPISerializer.extend({
             documentHash.meta.total_pages = Math.ceil(documentHash.meta.total / documentHash.meta.per_page);
         }
         return documentHash;
+    },
+
+    normalizeSingleResponse(_, __, ___, ____, requestType) {
+        const documentHash = this._super(...arguments);
+        if (requestType === 'findRecord' && documentHash.meta) {
+            documentHash.data.attributes.apiMeta = documentHash.meta;
```
adlius, line 149: can we do `documentHash.data.attributes.meta = documentHash.meta` without adding `apiMeta: DS.attr()` to the model? In fact, I'd expect every resource to have a `meta` property and am surprised that there isn't.

